Issue #7, opened 2020-05-03
Allow a custom stop-words list


jbesomi:
Allow a custom stop-words list to `remove_stop_words` function

**Edit** Starting from v1.0.6, the function is called `remove_stopwords` (without the second underscore)

Bronzekorean:
Hi Jonathan,

The project is very good and can save a lot of time for data exploration.

I've started working on this issue. I think the default behavior should still be to have the nltk stopwords.

I though of three ways to solve this: 
1- remove_stop_words(input: pd.Series, stop_words=stopwords.words("english"))
     raises two problems:
        - default is mutable
        - can't import without downloading the nltk stopwords before

2-  remove_stop_words(input: pd.Series, stop_words=None ) is less intuitive and requires adding a condition in the function core

3-  remove_stop_words(input: pd.Series, stop_words="english") have the stop_words be either a language to load from nltk or an iterable of strings


Which one do you think is best?

jbesomi:
Hi Bronzekorean, 

thank you for your positive replay and thank you for your help!

I agree with you we need to figure out a nice solution.

What do you think if we do like that:

Add a `get_default_stop_words()`
  - returns the defaukt NLTK stopwords
  - (addition) extra argument (`spacy=False`?) to decide whereas to return NLTK default words or SpaCy default words

Edit `remove_stop_words(input, stop_words=get_default_stop_words())`
- default stopwords is `get_default_stop_words()`
- In case a user wants a custom, it replaces the default with his own.
- It can also load `get_default_stop_words()` and append his own custom stop words.

ParthGandhi:
another way is to use a file with a namespaced attribute instead of a method i.e.

```python
from texthero import stop_words

remove_stopwords(input, stop_words.default)
remove_stopwords(input, stop_words.nltk)
remove_stopwords(input, stop_words.spacy_en)
remove_stopwords(input, None)
```

jbesomi:
It's also a good idea. Pick one and it's gonna be fine; as long as it's consistent.

PR #17, opened 2020-05-10
allow custom stop word lists

ParthGandhi:
fixes jbesomi/texthero#7


texthero/preprocessing.py
```diff
@@ -1,15 +1,22 @@
...
 from nltk.corpus import stopwords
-from nltk.stem import PorterStemmer
-from nltk.stem import SnowballStemmer
-import pandas as pd
-import numpy as np
+from nltk.stem import PorterStemmer, SnowballStemmer
+from spacy.lang.en import stop_words as spacy_en_stop_words
+
+
+class StopWordLists:
```
ParthGandhi, line 16: is this preferable to a separate `stopwords.py` file with top-level constants?
also might want to lazy load these vars if possible

jbesomi: Thank you for the request! Great.

Maybe is not necessary as there are just two stopwords constant, right? (spacy + nltk).

I will integrate the pull request shortly, first I need to merge the `test_preprocessing` as now it uses `unittest`.

ParthGandhi: in which case i think its preferable to have them as top-level constants instead of class vars. Lemme know if you disagree, otherwise will make the change.

Have you considered `pytest` instead of `unittest`? Also what was the reason to choose `yapf` over [black](https://github.com/psf/black)?

jbesomi: Agree on top-level constants.

`pytest` vs `unittest`: I just picked `unittest` without spending too much time deciding. Do you think we should switch to pytest? I'm open to other proposals as long there is a concrete difference.

`black` vs `yapf`. I picked `yapf` as the programmer does not have to make many decisions regarding code formatting. Looks like black is about the same. Do you think we should switch to black? I'm fine with that as well in case there is any valuable reason to chose it.

ParthGandhi:
@jbesomi updated. 

`pytest` is just a lot more modern and user friendly. There are several resources to see the differences ([example](https://www.slant.co/versus/9148/9149/~unittest_vs_pytest))

`black` is similar to `yapf` but is `psf` backed and likely what the community will standardize on. Lots of big projects use it: https://github.com/psf/black#used-by.

PR #18, opened 2020-05-10
Initial commit to support world-clouds visualization

### LINKED ISSUE #12: Wordclouds
Add wordclouds under visualization.

selimelawwa:
closes #12 
Basic implementation for word-clouds visualization
Add wordcloud to list of  in "requirements.txt" file

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ nltk
 scikit-learn
 plotly_express
 spacy
+wordcloud
--- a/texthero/visualization.py
+++ b/texthero/visualization.py
@@ -4,7 +4,7 @@ Visualize insights and statistics of a text-based Pandas DataFrame.
 
 import pandas as pd
 import plotly.express as px
-
+from wordcloud import WordCloud
 
 def scatterplot(df: pd.DataFrame,
                 col: str,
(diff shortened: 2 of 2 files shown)


jbesomi:
Thank you Selim!

It may be opportune to have as input a `pd.Series` rather than a `pd.Dataframe`. 

Also, for joining the text we may want to use:

```python
>>> import pandas as pd
>>> s = pd.Series(["a", "b", "c"])
>>> s.str.cat(sep=' ')
'a b c'
```

Can you also add some unittest? You can look at the official [world_cloud](https://github.com/amueller/word_cloud/tree/master/test) tests.

selimelawwa:
I have changed the function have as input a pd.Series @jbesomi .

Regarding unit tests. I have checked official [world_cloud](https://github.com/amueller/word_cloud/blob/master/test/test_wordcloud.py) tests. However current implementation of the word cloud method would not allow for implementing similar tests. Maybe we can have an underlining method _get_world_cloud which gets the WordCloud object which we can then test for words ..etc?

Do we only need to test the words in the generated word cloud? What else would you want to test?

jbesomi:
Thank you. Closing for now.

PR #21, opened 2020-05-21
Test case to remove brackets and has content

selimelawwa:


jbesomi:
Thank you. 

In `remove_whitespace`, why do you need to decode with `unicode_escape`? What does it happen if you don't or if you use the standard `utf-8` for instance?

In case we really need to stick to `unicode_escape`, we need to add a unittest that verify this so that we will not encounter again this problem in the future as the unittest will fail.

selimelawwa:
@jbesomi without `unicode_escape` remove white space method doesn't remove an escaped new line or tab i.e "\t\\n" will be "\n". Please check the screenshot example. I recognized this while testing has_content as you stated in its doc that an input with value "\t\\n" should be False

image

jbesomi:
Thank you Selim, that was insightful.

By default, with texthero, all input strings should not be escaped and therefore we should not call 'unicode_escape'. 

A string input "\t\n" means "tab" + "newline". This is clearly empty, i.e has _no_ content.

A string input "\t\\n" means "tab" + "\" + "newline". This is NOT empty, there is the backslash symbol, i.e _has_ content.

Now, the problem was that with the docstring, by default the strings are encoded and that's why there was "\t\\n". This is very confusing. I fixed that by defining the whole docstring as a "raw literal string" by prepending a 'r' to the docstring, i.e r"""this is a raw literal docstring.""""

In case of the unittest, `has_content(pd.Series("\t\n"))` should return  False, whereas  `has_content(pd.Series("\t\\n"))` should return True. I.e the function code is correct as it is.

jbesomi:
thanks.

PR #23, opened 2020-06-01
Remove unused import

Peritract:
The line `from nltk import NLTKWordTokenizer` in `visualization.py` causes import issues. I think it should probably be `from nltk.tokenize import word_tokenize`, but at present the import is not used in the file at all, so we can just remove it for simplicity's sake.


jbesomi:
Thanks Petriact!

PR #27, opened 2020-06-04
Spell checker

### LINKED ISSUE #14: Correct mistakes
Or at least check how many mistakes in a sentence.

See: https://pypi.org/project/pyenchant/

selimelawwa:
First implementation of correct_mistakes and correct_spacing methods
Added unit tests for both
Implementation using [symspellpy](https://github.com/mammothb/symspellpy)

closes #14

### DIFF
--- a/texthero/preprocessing.py
+++ b/texthero/preprocessing.py
@@ -15,6 +15,8 @@ from texthero import stopwords as _stopwords
 
 from typing import List, Callable
 
+import pkg_resources
+from symspellpy import SymSpell, Verbosity
 
 def fillna(input: pd.Series) -> pd.Series:
     """Replace not assigned values with empty spaces."""
@@ -614,3 +616,32 @@ def remove_urls(s: pd.Series) -> pd.Series:
     """
 
     return replace_urls(s, " ")
+
+
+def correct_mistakes(s: pd.Series) -> pd.Series:
+    r""" Correct spelling mistakes
+
+    `correct_mistakes` finds all spelling mistakes from the given Pandas Series and replace them with suggested words.
+
+    Examples
+    --------
+
+    >>> import pandas as pd
+    >>> import texthero as hero
+    >>> s = pd.Series(["whereis th elove hehad dated forImuch of thepast who couqdn'tread in sixtgrade and ins pired him"])
+    >>> hero.correct_mistakes(s)
+    0    where is the love he had dated for much of the past who couldn't read in six grade and inspired him
+    dtype: object
+    
+    """
+    sym_spell = SymSpell(max_dictionary_edit_distance=2, prefix_length=7)
+    dictionary_path = pkg_resources.resource_filename("symspellpy", "frequency_dictionary_en_82_765.txt")
+    bigram_path = pkg_resources.resource_filename("symspellpy", "frequency_bigramdictionary_en_243_342.txt")
+
+    sym_spell.load_dictionary(dictionary_path, term_index=0, count_index=1)
+    sym_spell.load_bigram_dictionary(bigram_path, term_index=0, count_index=2)
+
+    return s.apply(_correct_mistakes, args=(sym_spell,))
+
+def _correct_mistakes(text: str, sym_spell: SymSpell) -> str:
+    return sym_spell.lookup_compound(text, max_edit_distance=2)[0].term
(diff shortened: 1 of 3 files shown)


jbesomi:
Impressive work, thank you! Is there a situation where we want to correct the spacing but not the mistakes? `correct_spacing ` might be redundant, isn't?

Also, without knowing the library I would say that "spacing mistakes" are still mistakes and therefore as default `correct_mistakes` should take care of spacing mistakes. Opinions?

texthero/preprocessing.py
```diff
@@ -614,3 +616,54 @@ def remove_urls(s: pd.Series) -> pd.Series:
...
+    
+    """
+    sym_spell = SymSpell(max_dictionary_edit_distance=3, prefix_length=7)
+    dictionary_path = pkg_resources.resource_filename("symspellpy", "frequency_dictionary_en_82_765.txt")
+    bigram_path = pkg_resources.resource_filename("symspellpy", "frequency_bigramdictionary_en_243_342.txt")
+
+    sym_spell.load_dictionary(dictionary_path, term_index=0, count_index=1)
+    sym_spell.load_bigram_dictionary(bigram_path, term_index=0, count_index=2)
+
+    if fix_spacing is True:
```
mk2510, line 660: I think the is in python does just compare if two objects are identical. This won't fail here, since True is a basic type, but just `if fix_spacing:` would be more elegant

texthero/preprocessing.py
```diff
@@ -614,3 +616,54 @@ def remove_urls(s: pd.Series) -> pd.Series:
...
+
+    `correct_mistakes` finds all spelling mistakes from the given Pandas Series and replace them with suggested words.
+
+    When `fix_spacing` is set to ´True´, word_segmentation is applied.
+
+    Parameters
+    ----------
+    s : Pandas Series
+    fix_spacing : bool
+        When set to True, word_segmentation is applied, spacing is fixed in noisy text. Only use it if your data is similar to 'thequickbrownfoxjumpsoverthelazydog' as it slows performance
```
mk2510, line 632: it would be great, if the docstring length is no longer than 88 chars

texthero/preprocessing.py
```diff
@@ -614,3 +616,54 @@ def remove_urls(s: pd.Series) -> pd.Series:
...
+    sym_spell.load_dictionary(dictionary_path, term_index=0, count_index=1)
+    sym_spell.load_bigram_dictionary(bigram_path, term_index=0, count_index=2)
+
+    if fix_spacing is True:
+        s = s.apply(_correct_word_spacing, args=(sym_spell,))
+    
+    return s.apply(_correct_mistakes, args=(sym_spell,))
+
+def _correct_mistakes(text: str, sym_spell: SymSpell) -> str:
+    return sym_spell.lookup_compound(text, max_edit_distance=3)[0].term
```
mk2510, line 666: does lookup_compound always returns a value, because it looks quite dangerous to access the first entry in a list, without checking anything 🙈

texthero/preprocessing.py
selimelawwa: Ok will split it by new lines

texthero/preprocessing.py
selimelawwa: I think it should always return a value, but you are right, I  will recheck and if maybe add an if condition if needed

selimelawwa:
@mk2510 Can you please check what is failing in the build? I have resolved the merge conflicts and did the comments you requested. Unit tests are all succeeding.

Do you have any idea why build is failing?

Also it would be nice if we have a small chat today me, you and @jbesomi. We can have a short zoom call?

mk2510:
@selimelawwa I am sorry, I saw your push so late. Would tomorrow evening CET suit you as well? 
It looks like you forgot to format your files with the format.sh script, which you can find in the script folder :octocat:

jbesomi:
Hey guys, 

thank you Max for your valuable feedback!

Regarding the PR overall, I'm not sure whereas we really need this feature right now ... it reminds me of #129 ... the overall objective of Texthero since this respective issue has been open (#14, 08.05.2020) changed a bit...

What's your opinion Max the introduction of this new feature? Shouldn't we work all together focusing on the API checklist (#85)?

selimelawwa:
@mk2510 @jbesomi Ok we can have a call at like 6-7 pm today what do you think?

jbesomi:
I can't today at 18. What about tomorrow at 18:00?

selimelawwa:
Fine tomorrow at 18h

Issue #26, opened 2020-06-04
Add a flag to remove_punctuation to prevent removing punctuation in a token


jbesomi:
**Overview**

During pre-processing, when we need to remove punctuation, sometimes we want to preserve punctuation in the token. Example:

```
spider-man is powerful, isn't?
```

In this case, we might expect `remove_punctuation` to return:

```
spider-main is powerful isn't
``` 

**Approach**

Need to modify `remove_punctuation` and add a new argument `keep_tokens` or `remove_in_between` or something like that.

For the implementation, we can either tokenize the text (see `texthero.preprocessing.tokenize`) and remove all "punctuation tokens" or, probably better, add a regex that drops all punctuations symbols that are not between two characters (see again the tokenize function for an example of such regular expression).

**Open question**

Decide what does the default behaviour looks like. Probably, it's better to remove _all_ punctuation as default but make it clear that there is the opportunity to keep punctuation present in the tokens.

AdamHodgson:
I'd agree that the default behaviour should be to remove _all_ punctuation. 
I'll try to submit a pull request with these changes later today.

jbesomi:
Hi @AdamHodgson! Thank you. When you submit a PR, please do not forget to add new unit tests.

PR #37, opened 2020-07-08
Add a flag to remove_punctuation to prevent removing punctuation in a token

AdamHodgson:


jbesomi:
Hi @AdamHodgson, thank you for your PR! Amazing job!

It looks like your PR does not only change the code but also the documentation. Actually, this is my fault as it's not well explained on the CONTRIBUTING document; I will update that soon.

Can you please update your PR and just select the two files you effectively changed, namely `preprocessing.py` and `test/preprocessing.py`?

Once this new feature will be implemented in the new version (Texthero 1.1.0) I will then update the website.

AdamHodgson:
@jbesomi Please see the updated PR - I wasn't able to figure out how to link this change with [issue #26.](https://github.com/jbesomi/texthero/issues/26) 
For future reference, are you able to explain how I can do this? Thanks!

jbesomi:
Thanks!

You can read about [how to link a pull request to an issue there](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue). Thank you for asking, just learned something new :) Actually, is pretty easy, on the right-bar there is the "Linked Issue" tab; you just need to select it and pick the right issue.

As working with a regular expression and preprocessing is always tedious, for this PR I prefer to be a bit cautious. But we are gonna make it 👍 

Two comments:

1. If you look at the original code, all `remove_x` functions call `replace_x`. So, we should stick to the same paradigm.
1. By doing so you will also have to update `replace_punctuation` as well.
1. We need to do more tests to verify all possible edge cases, especially for `replace_punctuation`. You can look at the other tests under `test/preprocessing.py` for some ideas.

If everything is unclear, let me know!

AdamHodgson:
@jbesomi Please see the updated PR - hopefully this is in line with expectations. 

As requested, I've added some more tests to the `replace_punctuation` and `remove_punctuation` methods.

During testing, I noted that the existing methods handled blocks of punctuation (...) as single occurrences, so have updated the `keep_tokens` implementation to be consistent with the existing code.

Let me know if you'd like any further changes

jbesomi:
Great job @AdamHodgson ! Getting closer :)

Some feedback:
1. Great you checked the unchanged sentence and consecutive_punctuation 🥇 
1. You are still failing the Travis CI test. Did you `./format.sh` the code before pushing?
1. Having multiple unit tests make sense as long as we truly test for different scenario. Some of the unit tests does not really bring something new: 
 - check empty string: not sure is necessary
 - check for the unchanged sentence: ok, but one word is enough
-  replaces_all_punctuation: ok, but keep the text shorter
1. ` <PUNCT> ` is quite verbose and not strictly necessary. We might stick to something shorter, `X` should do the job. For future work; I encourage you to look at the already present code (look for instance under `replace_digits`)
1. For unit-tests; keep the sentence short. The idea is that we can quickly understand what we want to test; no need to add anything more. Also, keep the function names shorter (look at the other examples)
1. We need to check the edge-cases such as `falcon9`, `F.B.I`, ...
1. As we know that `remove_punctuation` just call `replace_punctuation` one simple test for each combination of arguments should be enough, right?

Extra:
I'm still not sure `keep_token` is the best name we can give to the argument ... but honestly I'm not really sure what is a better and valid alternative ... maybe `split_tokens`? Probably I would add the `s` anyway. What are your thoughts on this?

I just noticed that we are lacking some other tests for other functions, for instance `replace_stopwords`. We will need to consider also that (clearly not in this PR, but just for us to know) 

**Resources**:

If this is one of your first time doing unit-test, I encourage you to read a bit about the best practice in internet. Especially, I encourage you to read this article: [Testing Your Code](https://docs.python-guide.org/writing/tests/)

AdamHodgson:
Please see the updated PR 

Thanks for taking the time to help get this right!

jbesomi:
Hi Adam! Thank you, great job! I Will be back with you tomorrow (in about 16 hours, thank you for your patience, need to leave now)

jbesomi:
Hi Adam, sorry for the delay!

My only concern is that this code does not works correclty for some edge cases, for instance:

```
s = pd.Series("I live in U.K.")
hero.replace_punctuation(s, "X",keep_tokens=True)
I live in U.KX
```

This is not what the users want ... to solve this, though, we probably cannot rely to a simple regex formula (in general we want to remove `.` at the end of a sentence ...). This is not your fault, as the same problem exists also in the tokenization function. A solution would be to replace, both in the `tokenization` function as well as there (and probably in other places too), the regex formula with the spaCy tokenizer ... what's your opinion?

AdamHodgson:
Hi Jonathan, hope you're well 

I'm not certain I understand the problem. 

The example you have provided appears shows the same behaviour described in Issue #26.
For example: `'spider-man?'` becomes `'spider-man'` when preserving in-token punctuation.

I had assumed that the same would be true of your counterexample, I.E: `'U.K.' => 'U.K'`
Please could you help me understand why this is not what the users want?

Thanks

jbesomi:
Hi Adam,

thank you for your replay. I see what you mean and 100% agree with you that it makes sense to change the code only after having well understood the problem. My apologies also for having labeled the task as `good first issue`, it ends up more tricky than what I thought ...

In general, `U.K.` should be treated as a single token and that's why we would expect there that both dots would not be removed ..., you see?

If you are okay with that, we can put the PR as a draft and focus on other high-priority issues.

jbesomi:
Adam, I put it as a `draft` instead of `closed` as soon or later we will have to integrate it as this is super useful.

A follow-up question is the following:

1. We might want to requires as the input of the function an already-tokenized Pandas Series. That way, the function would remove only tokens composed of _only_ numeric values. If we do that, then we would have to do this for all `remove_x` and `replace_x` functions. What's your idea regarding that?

PR #39, opened 2020-07-08
Add black to be part of the CI

bobfang1992:
As per title so we enforce the style and never forget to black it.


jbesomi:
That's great! Thank you @bobfang1992 !

PR #51, opened 2020-07-09
Add count_sentences function to nlp.py

henrifroese:
Also add tests for the function to test_nlp.py
The function is implemented with spaCy, similarly to the other functions in the nlp module.
In response to #46

### DIFF
--- a/tests/test_nlp.py
+++ b/tests/test_nlp.py
@@ -57,5 +57,14 @@ class TestNLP(PandasTestCase):
 
     def test_count_sentences_index(self):
         s = pd.Series(["Test"], index=[5])
-        s_true = pd.Series([1], index=[5])
-        self.assertEqual(nlp.count_sentences(s), s_true)
+        counted_sentences_s = nlp.count_sentences(s)
+        t_same_index = pd.Series([""], index=[5])
+
+        self.assertTrue(counted_sentences_s.index.equals(t_same_index.index))
+
+    def test_count_sentences_wrong_index(self):
+        s = pd.Series(["Test", "Test"], index=[5, 6])
+        counted_sentences_s = nlp.count_sentences(s)
+        t_different_index = pd.Series(["", ""], index=[5, 7])
+
+        self.assertFalse(counted_sentences_s.index.equals(t_different_index.index))
--- a/tests/test_nlp.py
+++ b/tests/test_nlp.py
@@ -36,3 +36,16 @@ class TestNLP(PandasTestCase):
             [[("Today", "NP", 0, 5), ("such a beautiful day", "NP", 9, 29)]]
         )
         self.assertEqual(nlp.noun_chunks(s), s_true)
+
+    """
+    Number of sentences.
+    """
+
+    def test_number_sentences(self):
+        s = pd.Series("I think ... it counts correctly. Doesn't it? Great!")
+        s_true = pd.Series(3)
+        self.assertEqual(nlp.count_sentences(s), s_true)
+
+    def test_number_sentences_numeric(self):
+        s = pd.Series([13.0, 42.0])
+        self.assertRaises(ValueError, nlp.count_sentences, s)
--- a/tests/test_nlp.py
+++ b/tests/test_nlp.py
@@ -1,4 +1,5 @@
 import pandas as pd
+import numpy as np
 from texthero import nlp
 
 from . import PandasTestCase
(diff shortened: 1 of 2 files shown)


jbesomi:
It looks super nice! (and cool!)

This is definitely useful (Texthero's users will like it)

Some small changes / comments:

1. You did a great job! 👍 
1. For consistency, on `tests/test_nlp.py` "Number of sentences." should be replace with "Count sentences" and the same for the test functions names.
1. Great that you did "return pd.Series(..., `s=s.index`)". Forgetting `s.index` is a common mistake and it leads to problems. 
---> We need to test this; on _every_ Texthero's function (this will take a while ...). Can you please add a unit test for this case? You just need to define as input a Pandas Series with index not reset. 
---> Then, we will have to open a new issue and mention that we should test this on all Texthero's function that receives as input a Series and return a Series (almost all). You want to do it? I can review it later if you wish. Probably, we can add an extra `test_x` (name to be decided) file on the `tests` folder that we text for this kind of thing. 
1. Top that you select in the pipeline only the `sentencizer`
1. `s.astype("unicode").values` I forgot why we should add this, can you explain it? If we just pass `s.values`, what does it happen? [this might help](https://stackoverflow.com/questions/48177573/pandas-convert-object-column-to-str-column-contains-unicode-float-etc/49801130)
---> Also there we need to add unit-test to check this edge-cases
1. "Return a new Pandas Series with the results." is not very informative. Also, the body of the docstring and the head title should be separated by an empty line. Have a look at the [numpydoc docstring standard](https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard) section for hints.
1. `pd.api.types.is_string_dtype`, as we might to use this more than in one place, maybe `from pd.api.types import is_string_dtype` is a valid alternative.
---> Great catch anyway! We will need to add this kind of condition to other functions as well (will open an issue soon).

hf2000510:
Thanks, I've implemented the suggestions. I'll open the issue mentioned in 3.. About 5, I thought about it, and python already supports unicode in normal strings, so we actually do not need this. It rather is dangerous (it's also used in some other functions like noun_chunks) as e.g. we have

```
import pandas as pd
import numpy as np
s = pd.Series([np.nan])
s.values.astype('unicode')[0]
>> "nan"
```
so it converts the np.nan object that's used internally for missing values to the literal string "nan". This is certainly not what the user wants/expects.

Using pd.api.types.is_string_dtype(s) will not work, as it shows the following behaviour:
```
import pandas as pd
import numpy as np
s = pd.Series(["Test", 0])
t = pd.Series(["Test", np.nan])
pd.api.types.is_string_dtype(s)
>> True
pd.api.types.is_string_dtype(t)
>> True
 ```

Something like this
```
def _check_series_strings(s):
if not df.map(type).eq(str).all():
    raise TypeError("Non-string values in series. Use hero.drop_no_content(s) to drop those values.")
```
should work and is very fast.

So in total,
- changes to count_sentences are implemented. This does _not_ use .astype('unicode') anymore but instead lets the exception be thrown by spaCy for the time being (it's already correctly a TypeError, it will just be nicer if texthero throws it). Another issue/PR (next bullet point) will overhaul this for all functions where it makes sense.
- I'll open an issue to implement a function like above to check for strings and call this function where it makes sense.
- I'll open an issue to add tests for the index.

jbesomi:
Very good job and a very nice catch! 

By the way, what's your name? I suggest to add a photo profile on your Github profile and also have a more readable name ... it might be good pub for you!


1. Thank you for opening the issue #54, amazing. 
1. We might add a "See also" on the docstring where we reference the spacy `sentencizer`. This is a bit tricky, but a good learning :) you will need to edit intersphinx_mapping  dict of `conf.py` of the Sphinx configuration file (look under docs). For more info, you can refer to [sphinx.ext.intersphinx](https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html). Have you ever used Sphinx before? At the beginning is quite hard to use as the documentation is not particularly readable, but after a while, things will start making sense :). You will need to check that by running the website locally, again, some good learning 👍 
1. Is the term `item` the most appropriate? I never know how to precisely define a single cell of a Pandas Series ...

**How Texthero's should throw errors**

I opened an issue regarding that, please have a look at #60. We can either deal with that afterward this PR and merge it or again put it in `draft` mode and fix (?) #60 first ...

henrifroese:
> We might add a "See also" on the docstring where we reference the spacy sentencizer.

I couldn't yet get the website to work locally with the updated documentation (running the website works, but the script update_documentation.sh gives me errors - maybe that's not the correct way to update the documentation or I need to do something else first?), so I added a direct link to the spacy sentencizer documentation that will render as hyperlink in sphinx (should have the same effect). 

> I opened an issue regarding that, please have a look at #60. We can either deal with that afterward this PR and merge it or again put it in draft mode and fix (?) #60 first ...

I'd suggest not putting this PR on hold as #60 could take quite some time until everything surrounding that is finished. When implementing #60 and related issues like #55 most functions will need to be overhauled anyway.

> Is the term item the most appropriate?

I've changed it to "cell", that should be intuitive and I saw it's already used in some other locations in texthero.

Also, I implemented this :nerd_face:  :

> I suggest to add a photo profile on your Github profile and also have a more readable name

jbesomi:
Thank you!

All right, we will change it later on all functions. 

Regarding spaCy, we should do a comparison study and see if effectively using `spacy pipe` is faster than `pandas apply`. We will need to test this on a large dataset. Also, `batch_size` to 32 has been chosen quite randomly, we will need to test different values (or even approaches for that). I will open an issue now. 

You can now add another PR where you update the README by adding your name under contributors! 🎉🎉 congrats and welcome!

PR #57, opened 2020-07-09
Adding the code for POS tags

samirghouri:
Added the hero.pos in the nlp module.


samirghouri:
Hey @jbesomi , Can you please check this PR?

jbesomi:
Hi @samirghouri! Thanks a lot for your review and welcome to Texthero! 🎉

If you haven't already done, please read throughout [CONTRIBUTING.md](https://github.com/jbesomi/texthero/blob/master/CONTRIBUTING.md).

Your PR is also modifying the documentation. This should be avoided as this is not part of the POS. We will make the change on the website only after the POS function will be integrated into Texthero, starting from the next version. If you feel like the CONTRIBUTING.md document wasn't enough clear on that point, do not hesitate to improve it and open another PR.

**Some comments:**

1. For the POS, do we need both the `tagger` and `parser`? why is that? (you might need to refer to the [spaCy](https://spacy.io/usage/processing-pipelines#pipelines) )
1. As the contributing document explains, Texthero's workhorse is the quality of the documentation. We need to make sure the docstring is cristal clean. Have a look at [texthero.nlp.named_entities](https://texthero.org/docs/api/texthero.nlp.named_entities) for an idea (look also at the source code). Improvements consist in: 
 - Having a better heading title "Return a Pandas Series with part-of-speech tagging"
 - Show a list of possible POS tags as well as what they mean
 - Link to other resources (for example with the See Also section in the docstring). Refer to the numpy docstring guide for that (see CONTRIBUTING)

**Extra Improvements:**

1. `batch_size=32`. It would be great if you can do some research and understand which values we should use there. Even better, would be to test it on a concrete example. An alternative would also be to use `apply` instead of the for-cycle (`spacy pipe` shuld be multi-threading and that's why I picked this option but making a more in-depth analysis might be useful as-well) It's something you might be interested in? You can start by opening an issue for instance and says that we should focus more on that part. Thanks!

Regards,

jbesomi:
Hey @samirghouri, any news regarding that? :)

jbesomi:
I'm closing this as there hasn't been any update recently

PR #58, opened 2020-07-09
added replace hashtags and remove hashtag

ishanarora04:


ishanarora04:
Hey, Can you check the PR ?

jbesomi:
Hey @ishanarora04, it looks awesome!

Two very small improvements we can make:

- As we write `Example: #texthero_123.`, we can have the same hashtags example in the Example. I.e we can replace "Hi #hashtag, we will remove you." with "Hi #texthero_123, we will remove you.". Consistency is very useful when reading documentation.
- See also: "for replacing a tag with a custom symbol." should be "for replacing a _hashtag_ with a custom symbol."

Regards,

ishanarora04:
Done. Will ensure a full fledged description going forward.

jbesomi:
Thank you! Why on the `replace_tags` the pattern had round brackets `r"@([a-zA-Z0-9]+)"` whereas in `replace_hashtags` there are not? `"#[a-zA-Z0-9_]+"`.
Also, why you don't specify the `r` in the second case? Details matter
Can you please make a small research and argument your decisions? Sorry for being picky, just want to ensure we do not introduce bugs. Also, probably in the rest of the code we might have similar discrepancies. It would be great to fix all of this issues in case (in a separate PR).

By the way, how good is your knowledge of regular expression? If you need a recap, this is a super great website: [https://www.regular-expressions.info/](https://www.regular-expressions.info/)

Regards,

ishanarora04:
r is used to ensure a given string is treated as raw string. https://stackoverflow.com/questions/33729045/what-does-an-r-represent-before-a-string-in-python
The brackets are for the capturing group. Both the forms will provide with the desired effect.
However for our purpose 2nd form is better. I can provide a small PR asap to fix it in the tag form.

jbesomi:
Perfect, thank you! 🎉🎉

Feedback: when you post links on a comment when possible make use of the [markdown link tags](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links), it will be easier for others to read and understand!
  
Looking forward to see your next PR, you are learning soo fast!

PR #64, opened 2020-07-11
Preprocessing removed the capturing group from regex

ishanarora04:
We removed the capturing brackets from tags. The initial commit had no issues. Just for more clarity and adhering to convention.


jbesomi:
Thank you!

PR #66, opened 2020-07-11
Add tests for series indexes. Closes #54

### LINKED ISSUE #54: Add tests for (and implement when missing) using the input's index when returning a new Series.
For functions that take as input a pandas Series s, and then return a new pandas Series t, we have to be careful to use the index from s for the series t as well. This allows users to seamlessly integrate the function's result and continue working with it.

An example for a test would be the following (note it does not test whether the function works correctly, other test cases should check that; it only checks the index):
```
def test_count_sentences_index(self):
    s = pd.Series(["Test"], index=[5])
    counted_sentences_s = nlp.count_sentences(s)
    t_same_index = pd.Series([""], index=[5])

    self.assertTrue(counted_sentences_s.index.equals(t_same_index.index))

def test_count_sentences_wrong_index(self):
    s = pd.Series(["Test", "Test"], index=[5, 6])
    counted_sentences_s = nlp.count_sentences(s)
    t_different_index = pd.Series(["", ""], index=[5, 7])

    self.assertFalse(counted_sentences_s.index.equals(t_different_index.index))
```
This makes sure that the index returned by the function nlp.count_sentences is the same as the index in the input.

The tests could go into an extra file in the tests folder, e.g. test_indexes

henrifroese:
Tests are in tests/test_indexes.py

Fully parameterized to allow for easy expansion.

Library "parameterized" added as import for this.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -38,7 +38,6 @@ install_requires =
     unidecode>=1.1.1
     gensim>=3.6.0
     matplotlib>=3.1.0
-    parameterized>=0.7.4
 # TODO pick the correct version.
 [options.extras_require]
 dev =
@@ -48,3 +47,4 @@ dev =
     sphinx-markdown-builder>=0.5.4
     recommonmark>=0.6.0
     nbsphinx
+    parameterized>=0.7.4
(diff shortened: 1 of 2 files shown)


henrifroese:
I've now switched the "parameterized" dependency to dev only, makes sense! Also implemented the tests for the other modules except from nlp, so now nlp+preprocessing+visualization+representation are fully covered regarding the indexes.

jbesomi:
Thanks; that's super good!

Issue #60, opened 2020-07-10
Kind of Pandas Series


jbesomi:
**Motivation**

Having a unified view and a clear idea of the expected Pandas Series input it's useful both for the users and for the developers.

To receive precise and correct _errors_ is very valuable for the users as this permits an easy and pleasant debugging. We can summarize three _kinds_ of Pandas Series a Texthero's function can receive as input (or it can output):

**Types**

- "Pandas Text Series" --> every `cell` has some text
- "Pandas Tokenized Series" --> every `cell` has a list of tokens 
- "Pandas Representation Series" --> every cell is a representation of a text ( it's a list of `float` values). This will be improved soon (See issue #43)

In the best scenario, every Texthero's function receive as input a Pandas Series of one of these three _kind_. Testing that the given Pandas Series is of the right expected types is therefore useful. 

**Go further**

- `preprocess.py`: almost all function (at the exception of `tokenize`) takes as input a Pandas Text Series and Return a Pandas Text Series.
- `represention.py`: input (will be #44 ) a Tokenized Pandas Series and output will be Representation Pandas Series
- `nlp.py`: input is a Text Pandas Series, whereas the output is _TODO_
- `visualization.py` _TODO_.

It would be great to have a unified and clear view of all this:

1. Every function should check for the right type (we will need to define the "check" function, probably under a new file, something like `_helper.py`)
1. Once everything is in place and defined, add under the website (documentation) a clear document that explain all this. It will be so easy to use Texthero then!
1. New ideas

**Extra**

Unfortunately, there are more variants of Pandas Series (output of `named_entities`, output of `pca`, ...) there is still some design work to go there ...

**Work in progress ...**

henrifroese:
I just opened a first draft PR in #69 as a step to implementing this. I'll copy the new file _helper.py 's docstring here:

Hero Series Types
=================

There are different kinds of Pandas Series used in the library, depending on use.
For example, the functions in preprocessing.py usually take as input a Series
where every cell is a string, and return as output a Series where every cell
is a string. To make handling the different types easier (and most importantly
intuitive for users), this file implements the types as subclasses of Pandas
Series and defines functions to check the types.

These are the implemented types:

- TextSeries: cells are text (i.e. strings), e.g. "Test"
- TokenSeries: cells are lists of tokens (i.e. lists of strings), e.g. ["word1", "word2"]
- RepresentationSeries: cells are vector representations of text (see issue #43), e.g. [0.25, 0.75]

You could now do this:
[...]

PR #69, opened 2020-07-11
Checking for Series Types (Representation Series, Tokenized Series, ...).

henrifroese:
This is mainly meant as a first step towards closing #60 and should not be merged yet. It implements three types of custom pandas Series that implement the most common usages in the library. Also adds decorators for easier development using the types. See the discussion in #60  for more information.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -74,3 +74,70 @@ class TestHelpers(PandasTestCase):
         with warnings.catch_warnings():
             warnings.simplefilter("ignore")
             self.assertTrue(f(s).index.equals(s_true.index))
+
+    """
+    InputSeries.
+    """
+
+    def test_inputseries_function_executes_correctly(self):
+        @_helper.InputSeries(_helper.TextSeries)
+        def f(s, t):
+            return t
+
+        s = pd.Series("I'm a TextSeries")
+        t = "test"
+        self.assertEqual(f(s, t), t)
+
+    def test_inputseries_wrong_type(self):
+        @_helper.InputSeries(_helper.TextSeries)
+        def f(s):
+            pass
+
+        self.assertRaises(TypeError, f, pd.Series([["token", "ized"]]))
+
+    def test_inputseries_correct_type_textseries(self):
+        @_helper.InputSeries(_helper.TextSeries)
+        def f(s):
+            pass
+
+        try:
+            f(pd.Series("I'm a TextSeries"))
+        except TypeError:
+            self.fail("Failed although input type is correct.")
+
+    def test_inputseries_correct_type_tokenseries(self):
+        @_helper.InputSeries(_helper.TokenSeries)
+        def f(s):
+            pass
+
+        try:
+            f(pd.Series([["token", "ized"]]))
+        except TypeError:
+            self.fail("Failed although input type is correct.")
+
+    def test_inputseries_correct_type_vectorseries(self):
+        @_helper.InputSeries(_helper.VectorSeries)
+        def f(s):
+            pass
+
+        try:
+            f(pd.Series([[0.0, 1.0]]))
+        except TypeError:
+            self.fail("Failed although input type is correct.")
+
+    def test_inputseries_correct_type_documentrepresentationseries(self):
+        @_helper.InputSeries(_helper.DocumentRepresentationSeries)
+        def f(s):
+            pass
+
+        try:
+            f(
+                pd.Series(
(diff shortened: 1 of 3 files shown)


jbesomi:
Henri you are awesome!! 🎉🎉

henrifroese:
We have now worked on this again; should be ready for a review/discussion (and maybe a merge) @jbesomi .

Summary:
- Everything is explained in `_helper.py`
- Classes `TextSeries, TokenSeries, VectorSeries, DocumentRepresentationSeries` where each class has a function that can check for a Series `s` if `s` fulfills the type
- Decorator `InputSeries(allowed_hero_series_type=)` to check the input series of a function
- Tests for the decorator

Uses:

We can now do two things:

```
>>> import pandas as pd
>>> from texthero._helper import *
>>> def tfidf(s: TokenSeries) -> DocumentRepresentationSeries:
>>>      # tfidf stuff
```
This is very expressive for the users in the documentation (and also for the developers)! Everyone can immediately see what tfidf is working on and what the output looks like. Doing `help(TokenSeries)` now also explains to the users what's going on.

This does *not* force users to do anything differently from before, they should just continue using Pandas Series as usual; it's mainly for documentation / ease of use / expressiveness. This also automatically adds good explainations through the docstrings.

Second of all, the decorator (probably self-explanatory):

```
>>> from texthero._helper import *
>>> import pandas as pd
>>> @InputSeries(TokenSeries)
... def f(s):
...     pass
>>> f(pd.Series("Not tokenized")) # doctest: +SKIP
>>> # throws a type error with a nice explaination
>>> f(pd.Series([["I", "am", "tokenized"]]))
>>> # passes
```

All of this is implemented very lightweight and adds basically no overhead; the type checking etc. is all `O(1)`

tests/test_helpers.py
```diff
@@ -74,3 +74,70 @@ def f(s):
...
+            pass
+
+        self.assertRaises(TypeError, f, pd.Series([["token", "ized"]]))
+
+    def test_inputseries_correct_type_textseries(self):
+        @_helper.InputSeries(_helper.TextSeries)
+        def f(s):
+            pass
+
+        try:
```
jbesomi, line 103: Why do you have the `try`/`except` there? Might be overcomplicated, isn't it?

texthero/_helper.py
```diff
@@ -70,3 +71,242 @@ def wrapper(*args, **kwargs):
...
+                return True
+
+        def is_list_of_numbers(cell):
+            return all(is_numeric(x) for x in cell) and isinstance(cell, (list, tuple))
+
+        if not is_list_of_numbers(s.iloc[0]) or s.index.nlevels != 1:
+            raise TypeError(error_string)
+
+
+class DocumentRepresentationSeries(HeroSeries):
```
jbesomi, line 258: I see why you called it `Document..` but would rather keep it more general: `RepresentationSeries`

Example: word embeddings. This is can also be seen as `RepresentationSeries`, the first index is the word and the second index is the `dim_x` (but it's also true that this can also be a `VectorSeries` ). Having a separate Series for this would not be necessary...

If we end-up seeing we want to distinguish between the twos, then we might create other two classes `DocRepresentationSeries` and `WordRepresentationSeries` and inherit `RepresentationSeries` ... but then we might also need `TokenRepresentationSeries` ... -> `RepresentationSeries` should be enough.

tests/test_helpers.py
henrifroese: We want to make sure that calling it with the right type does not raise an Exception here. I'm going off of [this stackoverflow discussion](https://stackoverflow.com/questions/4319825/python-unittest-opposite-of-assertraises). We could just do e.g. `f(pd.Series("I'm a TextSeries"))`, but if that raises an Exception, we get an _exception_ in the test suite which makes us think a test is written wrong, instead of correctly getting a _failure_  telling us "test is programmed correctly, but it's failing".

texthero/_helper.py
henrifroese: Makes sense, will do

henrifroese:
Just renamed "DocumentRepresentationSeries" to "RepresentationSeries"

tests/test_helpers.py
jbesomi: Thank you! Great job!

jbesomi:
Everything looks great! I will merge it now ... the only observation: what if we move this under a separate file? Like `types.py` or `heroseries.py` or something like that?

PR #74, opened 2020-07-12
Implement Automated Readability Index, Closes #20 ; new PR

### LINKED ISSUE #20: Automated readability index

henrifroese:
New pull request from #46 as we had some Git problems.

Input checking done with pd.api.types.is_string_dtype. Not a
permanent solution, will be improved by #60 etc.

### DIFF
--- a/tests/test_visualization.py
+++ b/tests/test_visualization.py
@@ -59,3 +59,25 @@ class TestVisualization(PandasTestCase):
     def test_wordcloud(self):
         s = pd.Series("one two three")
         self.assertEqual(visualization.wordcloud(s), None)
+
+    """
+    Test automated readability index
+    """
+
+    def test_automated_readability_index(self):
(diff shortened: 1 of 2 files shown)


jbesomi:
This looks very good!

The only question: "Score is NaN if it cannot be computed (e.g. if the number of sentences is 0).", are we testing this scenario there?

henrifroese:
You're right, I totally missed that. I just improved the testing at the beginning of the function in a new commit. I also added a test for that.

jbesomi:
Henri, probably this function will fail this test: #86. What if we put this aside as a draft for a moment, work on the rest, and come back to that later after having implemented #86 in a couple of weeks?

henrifroese:
Yes, makes sense

PR #90, opened 2020-07-15
Make Text-Based Representation functions take tokenized input. Closes #44

### LINKED ISSUE #44: TokenSeries as input to every representation function
One of the principles of Texthero is to give to the NLP developer more control.

**Motivation**

A simple example is the [TfidfVectorizer](https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.TfidfVectorizer.html) object from the scikit learn. It's absolutely fast and great but it has too many parameters and before applying the TF-IDF it actually preprocesses the text data. I just discovered that TfidfVectorizer even L2 normalizes the output and that there is no option to avoid a normalization.

 With Texthero's `tf-idf` we just want the code to apply TF-IDF. That's it. No stopwords removal, no tokenization, no normalization. All this essential step can be done by the NLP developer on the pipeline (the drawback is that it might be less efficient, but at the advantage of having clear and expected behavior). 

**Solution**

All `representation` functions will require the Pandas Series to be already tokenized. In the beginning, we can still accept Text Pandas Series; in this case the default `hero.tokenize` the function will be applied but a warning message will be outputted (see example below).

**Interested in working on this task?**
For the `tfidf` + `term_frequency` function, the code has already (almost) been made. The body of the function would look like this:

```
if type(s.iloc[0]) != list:
    raise ValueError(
        "🤔 It seems like the given Pandas Series is not tokenized. Have you tried passing the Series through `hero.tokenize(s)`?"
    )

tfidf = TfidfVectorizer(
    use_idf=True,
    max_features=max_features,
    min_df=min_df,
    max_df=max_df,
    tokenizer=lambda x: x,
    preprocessor=lambda x: x,
)
```

If you are interested in helping out, just leave a comment!

henrifroese:
Now checks for all text-based functions in representation.py whether the
input is already tokenized. If not,
it prints a warning and tokenizes.

Docstrings and tests are changed accordingly.

### DIFF
--- a/tests/test_indexes.py
+++ b/tests/test_indexes.py
@@ -55,8 +55,12 @@ test_cases_preprocessing = [
 ]
 
 test_cases_representation = [
-    ["term_frequency", representation.term_frequency, (s_text,)],
-    ["tfidf", representation.tfidf, (s_text,)],
+    [
+        "term_frequency",
+        representation.term_frequency,
+        (preprocessing.tokenize(s_text),),
+    ],
+    ["tfidf", representation.tfidf, (preprocessing.tokenize(s_text),)],
     ["pca", representation.pca, (s_numeric_lists, 0)],
     ["nmf", representation.nmf, (s_numeric_lists,)],
     ["tsne", representation.tsne, (s_numeric_lists,)],
--- a/tests/test_representation.py
+++ b/tests/test_representation.py
@@ -25,22 +25,31 @@ class TestRepresentation(PandasTestCase):
 
     def test_term_frequency_single_document(self):
         s = pd.Series("a b c c")
+        s = preprocessing.tokenize(s)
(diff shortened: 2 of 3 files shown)


jbesomi:
Nice, well done! 👍 

Review:

 1. The Word2Vec code is there by mistake. I will remove it soon and we will just use the "flair" (so that we avoid training, this is out-of-scope of the Texthero's purpose). Sorry about that!
 1. We probably do not need anymore `token_pattern="\S+"`, right? Again my fault!

henrifroese:
Thanks, yes, makes sense with the word2vec. I removed the token_pattern argument.

jbesomi:
Looks great! Can I merge it?

(I will then add 'min_features' in both places and remove 'lowercase=False' too)

henrifroese:
Yes, thanks!

PR #97, opened 2020-07-15
Improve tfidf

henrifroese:
Docstring now includes formula/explaination.

Normalization disabled (the option "normalization=None" was "hidden" in the sklearn code, so that turned out to be an easy fix).

Representation Series is already being handled (although output is still like before, using representation_series_to_flat_series).

Function representation_series_to_flat_series added.

Unit tests are changed accordingly, also one with the explicit calculation using the formula.


jbesomi:
Nice! 🎉 

Review:

1. Is said that `max_features` has default values `300` but is `None`.
2. `lowercase=False` not necessary as is already tokenized
3. We should probably mention in the docstring that in contraposition to scikit-learn TfidfVectorizer (add a link), `tfidf` do not normalize the ouput (we will later add under "See Also" `.norm()`)
4. If you want to go further, you might want to unit-test `max_features`, `min_df` and `max_df` 👍

henrifroese:
Thanks, fixed the issues (not without some git trouble :face_with_head_bandage:). 

> If you want to go further, you might want to unit-test max_features, min_df and max_df +1

This lead to finding some other bugs in the functions, so thanks! Testing really does work :neckbeard:

jbesomi:
> Thanks, fixed the issues (not without some git trouble 🤕).

I see; take it as a learning experience ... when working with Git, something suuper useful is to have a software to deal with this kind of troubles; I love and use [Fork](https://git-fork.com/): it's fast and friendly. Check it out!

> This lead to finding some other bugs in the functions, so thanks! Testing really does work 

That's great! I'm also learning the importance of always testing! What were the issues and how did you find them? I'm sure I can learn something :)

Review:

Everything seems great! The only part I'm not 100% convinced is the `representation_series_to_flat_series ` function (but it's the hardest part), especially:
 - Would be better if we avoid the reindex as well as having to pass the flattened index as argument
 - What is called "missing values" are tfidf values for term that are not present in the document. The values, in this case, should rather be 0.

For you to know, I will test this part later today or tomorrow.

henrifroese:
> What were the issues and how did you find them? I'm sure I can learn something :)

For example, using the min_df parameter, I noticed that the documents for which no token was selected as a feature, they completely disappeared. So to be able to continue using the index, I had to change some stuff (see below).

> Would be better if we avoid the reindex as well as having to pass the flattened index as argument

For the ```representation_series_to_flat_series```, the function is meant to handle all Representation Series, not just the tfidf one, and in general I believe that we cannot recover the original Index from a Sparse Representation Series. For example, if we have a Pandas Series with documents doc1, doc2, doc3, doc4 in the index, if the tfidf (or another function later on) does not calculate any tfidf scores for doc3 (e.g. because all the words in doc3 are below the min_df threshold), the Sparse Pandas Series will just leave doc3 out completely. So to be sure to recover all documents, we need to pass the original index of s. That is what I noticed through testing with min_df. If no documents are "lost" in the Sparse Representation, we have no problem. But if it can be the case that we calculate no features for some documents, they would be lost instead of NaN without the reindex.

> What is called "missing values" are tfidf values for term that are not present in the document. The values, in this case, should rather be 0.

Yes, when  ```representation_series_to_flat_series``` is called in tfidf, it's called with fill_missing_values=0.0 to do this. But if there's a document where the actual value is nan before calling tfidf, it will stay which I think makes sense.

> I love and use Fork: it's fast and friendly. Check it out!

Thanks, I'll have a look!

jbesomi:
Thank you for pointing these details out; now it definitely makes more sense to me; sorry! You are the real hero there! 👍

jbesomi:
@mk2510 Max and @henrifroese Henri you did an awesome job! 🎉 🍾 !

PR #104, opened 2020-07-18
Updated scatterplot docstring, also top_words docstring. Closes #100

### LINKED ISSUE #100: Improvement in scatterplot's docstring.
In relation to the issue #98 

Explaining about the `color` argument for `numerical` and `categorical` data types in the `scatterplot` docstring

vidyap-xgboost:
In `scatterplot`, also made a change to the argument `return_figure`, by default it was set to `False`, but it still returns the figure, why? Because `fig.show()` was not commented out and it makes more sense to put the default as `True` because the function is built to return the `scatterplot`. Hence, setting `return_figure` as `True`.

Changed the grammar in `top_words` function. 

For `scatterplot` and `top_words`, I followed [`numpy docstring guide`](https://numpydoc.readthedocs.io/en/latest/format.html) as provided in the `CONTRIBUTING.md` and also an [`example`](https://numpydoc.readthedocs.io/en/latest/example.html#example) from `numpy docstring guide`.

Closes #100

### DIFF
--- a/texthero/visualization.py
+++ b/texthero/visualization.py
@@ -56,7 +56,7 @@ def scatterplot(
         df, x=pca0, y=pca1, color=color, hover_data=hover_data, title=title
     )
     # fig.show(config={'displayModeBar': False})
-    fig.show()
+    # fig.show()
 
     if return_figure:
         return fig
--- a/texthero/visualization.py
+++ b/texthero/visualization.py
@@ -22,18 +22,33 @@ def scatterplot(
     color: str = None,
     hover_data: [] = None,
     title="",
-    return_figure=False,
+    return_figure=True,
 ):
-    """
+    r"""
     Show scatterplot using python plotly scatter.
 
     Parameters
     ----------
-    df
-    col
+    df : pd.DataFrame
+    col : str
         The name of the column of the DataFrame used for x and y axis.
+    color : str, optional
+        Grouping variable that will produce points with different colors. Can be either numeric or or object or categorical. 
+        If `color` is a pd.Series with numerical values, color mapping is of a continuous color bar.
+        If `color` is an object or a pd.Series with categorical values, color mapping is of a discrete color bar.
+    hover_data : object, optional
+        A single hover label appears for the point directly when the cursor is placed on the point.
+    title : str, optional
+        Title of the scatterplot.
+    return_figure : bool, optional
+        If set to `True` (default), returns the scatterplot. If set to `False`, doesn't return the scatterplot.
+    
+    Returns
+    -------
+    scatterplot
+        If `return_figure` is set to True (default), returns the scatterplot.
+        If `return_figure` is set to False, doesn't return the scatterplot.
     """
-
     pca0 = df[col].apply(lambda x: x[0])
     pca1 = df[col].apply(lambda x: x[1])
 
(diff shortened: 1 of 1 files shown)


henrifroese:
Hi, I actually believe this is a bad idea. With this, users would have to visualize the returned plot for themselves. I think it's much nicer if

1. return_figure=True -> no plot is shown, figure is returned
2. return_figure=False -> plot is shown, figure is not returned

This is what we implemented in #107 , please have a look and maybe consider closing this if you agree.

Thanks,
Henri

vidyap-xgboost:
> Hi, I actually believe this is a bad idea. With this, users would have to visualize the returned plot for themselves. I think it's much nicer if
> 
> 1. return_figure=True -> no plot is shown, figure is returned
> 2. return_figure=False -> plot is shown, figure is not returned
> 
> This is what we implemented in #107 , please have a look and maybe consider closing this if you agree.
> 
> Thanks,
> Henri

Hi, I fail to understand how returning 2 plots can help the user. Could you explain a bit more? 

Thanks.

henrifroese:
Yes, I'll clarify:

We have two things:
1. a _return value_ . This is what the function returns to the user.
2. the plot/figure as _side effect_ that the users can actually see on the screen.

If we return the plot with `return fig`, the plot itself is not shown. If we use `fig.show()`, the plot is shown. The only reason why we might want to use `return fig` that I can think of is that a user might want to modify the plotly plot somehow.

Consider the following example:
```
>>> import plotly.express as px
>>> def f():
>>>     fig = px.scatter(x=[0, 1, 2, 3, 4], y=[0, 1, 4, 9, 16])
>>>     return fig
>>> f()  # try it out, does not show the plot!
Figure({
    'data': [{'hovertemplate': 'x=%{x}<br>y=%{y}<extra></extra>',
              'legendgroup': '',
              'marker': {'color': '#636efa', 'symbol': 'circle'},
              'mode': 'markers',
              'name': '',
              'showlegend': False,
              'type': 'scatter',
              'x': array([0, 1, 2, 3, 4]),
              'xaxis': 'x',
              'y': array([ 0,  1,  4,  9, 16]),
              'yaxis': 'y'}],
    'layout': {'legend': {'tracegroupgap': 0},
               'margin': {'t': 60},
               'template': '...',
               'xaxis': {'anchor': 'y', 'domain': [0.0, 1.0], 'title': {'text': 'x'}},
               'yaxis': {'anchor': 'x', 'domain': [0.0, 1.0], 'title': {'text': 'y'}}}
})
>>> 
>>> def g():
>>>     fig = px.scatter(x=[0, 1, 2, 3, 4], y=[0, 1, 4, 9, 16])
>>>     fig.show()
>>> g()  # try it out, this shows the plot!
```

So `fig.show()` shows the plot to the user and `return fig` returns the figure object to the user so he could change it if he wants to (probably unlikely, so default should be to just show the plot).

vidyap-xgboost:
@henrifroese Thank you for explaining! I understood now. Closing this pull request since you're already working on it.

jbesomi:
> @vidyap-xgboost @henrifroese @sleeper @cclauss @ParthGandhi #85 # #111 @ #107

Hey @lydz1209, what does that mean?

Issue #65, opened 2020-07-11
Make spaCy-nlp functions faster


jbesomi:
Almost all functions of the `nlp` module under-the-hoods make use of [spaCy](https://spacy.io/).

In general, spaCy is quite fast as it uses Cython.

The core code looks like this:

```
new_data = []
for row in nlp.pipe(s.values, batch_size=32):
        new_data.append( ... row ...)
```

`spacy pipe` has been initially chosen as it's [multi-threading](https://explosion.ai/blog/multithreading-with-cython). An alternative might be to use `apply` ( probably is slower). 

The `pipe` functions have among other the `n_threads` as well as the `batch_size` arguments. Tuning this values might be very important.

This task consists in:

1. Understand spaCy `pipe`
2. Test on a large dataset different combinations of `n_threads` and `batch_size` value
3. (it it make sense) Compare this results with the `pandas apply` approach
4. Pick the best solution and implement it in all NLP functions that uses spaCy under-the-hoods
  - We might find that the optimal values of `n_threads` and `batch_size` are not always the same, in this case, we will need to add it as arguments to the NLP functions and update the docstring.

ishanarora04:
Okay, I am onto this.

mk2510:
Hi Jonny,

we will now start with this issue and implement multi_processing, wherever it is useful. 🚀

jbesomi:
Hi Max,

I guess this task might take quite long time, what if we prioritize to completely finish part 2 of the "API next checklist" and then move on to part 4?

henrifroese:
Will do :2nd_place_medal:  :neckbeard:

jbesomi:
**Dask vs. spaCy**

It's faster to use `pipe` from spaCy or to directly use `Dask` ([Dask Dataframe](https://docs.dask.org/en/latest/dataframe.html))?

`Dask` might be the solution we were looking for ...

PR #106, opened 2020-07-18
Added the function to  POS tag

Rickecr:
Added the function to  POS tag.

I saw the suggestions in PR #57 and read the CONTRIBUTING.md file.

Any suggestions please tell me. Thank you!

### DIFF
--- a/tests/test_nlp.py
+++ b/tests/test_nlp.py
@@ -79,20 +79,20 @@ class TestNLP(PandasTestCase):
         s_true = pd.Series(
             [
                 [
-                    ("Today", "NOUN", "NN"),
-                    ("is", "AUX", "VBZ"),
-                    ("such", "DET", "PDT"),
-                    ("a", "DET", "DT"),
-                    ("beautiful", "ADJ", "JJ"),
-                    ("day", "NOUN", "NN"),
+                    ("Today", "NOUN", "NN", 0, 5),
+                    ("is", "AUX", "VBZ", 6, 8),
+                    ("such", "DET", "PDT", 9, 13),
+                    ("a", "DET", "DT", 14, 15),
+                    ("beautiful", "ADJ", "JJ", 16, 25),
+                    ("day", "NOUN", "NN", 26, 29),
                 ],
                 [
-                    ("São", "PROPN", "NNP"),
-                    ("Paulo", "PROPN", "NNP"),
-                    ("is", "AUX", "VBZ"),
-                    ("a", "DET", "DT"),
-                    ("great", "ADJ", "JJ"),
-                    ("city", "NOUN", "NN"),
+                    ("São", "PROPN", "NNP", 0, 3),
+                    ("Paulo", "PROPN", "NNP", 4, 9),
+                    ("is", "AUX", "VBZ", 10, 12),
+                    ("a", "DET", "DT", 13, 14),
+                    ("great", "ADJ", "JJ", 15, 20),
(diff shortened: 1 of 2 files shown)


jbesomi:
Thank you @Rickecr, that's amazing!! 🎉 🎉 

What if we add also `starting char` and `ending char` as in [texthero.nlp.named_entities](https://texthero.org/docs/api/texthero.nlp.named_entities)? Can you see a value in adding this?

Regarding this comment in #57:

> `batch_size=32`. It would be great if you can do some research and understand which values we should use there. Even better, would be to test it on a concrete example. An alternative would also be to use apply instead of the for-cycle (spacy pipe shuld be multi-threading and that's why I picked this option but making a more in-depth analysis might be useful as-well) It's something you might be interested in? You can start by opening an issue for instance and says that we should focus more on that part. Thanks!

Did you have the occasion to study this problem further? The same question is also asked there as a separate issue #65. Any help with this too would be super appreciated 👌 

---

P.s I have been to Brazil some years ago, Florianopolis, Santa Caterina. I'm missing the vibe and my time there (muita saudade)!! You live in such a beautiful country!

Rickecr:
Hello @jbesomi.

1 - I think it's a good idea to put the `starting char` and `ending char`, because maybe it's important information and I also think it's good to follow the pattern of the functions. Can I add to this PR ?

2 - About the question I even read about it(for this [text](https://prrao87.github.io/blog/spacy/nlp/performance/2020/05/02/spacy-multiprocess.html#Effect-of-chunk-size-and-batch-size)), as I understand it depends on the size of the documents used, so I think it may be a good option to leave this as a parameter for the user, and so he could test with his data. What do you think ?

P.s Thanks for the compliments to my country, the energy here is really good. I hear very good things about your country, hope to visit there one day. And sorry for the bad english.

jbesomi:
1. All right!
2. Ok, but I guess we should do some test and evaluation to at least suggest to users valid values. We will deal with this in a separate PR.

Rickecr:
ready, I will add this change discussed in point 1. And we can deal with the issue of batch size, it is a good idea.

Rickecr:
I'm ready for the review

jbesomi:
Looks great!

Review:

1. The only part that is not 100% clear is the difference between `POS tag` and `The detailed POS tag`. Can you please add a sentence in the docstring explaining the difference between the two?
1. "List of labels:" rename to "List of POS/Tag"? (docstring must be clear, clean and precise)
1. Need to add under the docstring the missing list of POS/Tag 
1. To more details: `POS tagging <https://spacy.io/api/annotation#pos-tagging>` -> we want to provide to the users already all required information to use such function. At the same time though, we want to show where this solution comes from ...
Look at the other code, something like "Internally `noun_chunks` makes use of Spacy's dependency parsing:
    https://spacy.io/usage/linguistic-features#dependency-parse" might be more appropriate ...

Rickecr:
Hey @jbesomi . I made the requested changes, less the number 3, because I could not find a list of absence of POS / tag. I only found information about the POS/Tag list:

- List of POS/Tag: https://spacy.io/api/annotation#pos-universal
- Universal POS tags: https://universaldependencies.org/u/pos/

jbesomi:
Thank you @Rickecr, merged!

And officially welcome to Texthero's contributors! You can now open a new PR and add your name under README.md - contributors! 👍 🎉 🎉

PR #112, opened 2020-07-22
Initial commit to explain how to handle PDF date

### LINKED ISSUE #24: Support reading text data from PDF and PowerPoint
PDF, PowerPoint presentations and other unstructured text, contain very valuable data that can be used for analysis. 
There are many tools providing this features. It would be nice if we can provide a single method to read such files and don't bother user with this.

There is a python library [textract](https://textract.readthedocs.io/en/stable/index.html) provide this functionality unfortunately it is not maintained. 

We can provide a method loadData or so that has different implementation depending on file type

selimelawwa:
Initial commit on blog for handling PDF files. Please check and let me know your comments

Remaining:
- Example with: [pdfminer](https://github.com/pdfminer/pdfminer.six)
- Example with: [Apache Tika](https://github.com/chrismattmann/tika-python)
- Conclusion
- Maybe more examples with texthero

Resolves #24


jbesomi:
Thank you Selim! This is a very good start, keep going! ;)

PR #113, opened 2020-07-22
Update Docstrings in Preprocessing Module

henrifroese:
* Add examples, fix formatting, fix spelling mistakes, improve explainations

* Rename argument "input" to "s"

### DIFF
--- a/texthero/preprocessing.py
+++ b/texthero/preprocessing.py
@@ -100,11 +100,11 @@ def replace_digits(s: pd.Series, symbols: str = " ", only_blocks=True) -> pd.Ser
 
 def remove_digits(s: pd.Series, only_blocks=True) -> pd.Series:
     """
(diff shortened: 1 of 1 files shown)


jbesomi:
Amazing, thx! 🎉 

Review:

1. According to [PEP 257](https://www.python.org/dev/peps/pep-0257/):

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

So, "return" without the _s_ should be preferred there. Find-replace "Returns" -> "Return" and "Removes" -> "Remove" should do the job

henrifroese:
Makes sense, done

jbesomi:
Top! Thanks! Merged 👍

PR #116, opened 2020-07-22
Update docstrings in representation

henrifroese:
* All docstrings now have correct explainations and parameter descriptions

* Add some argument/output types

*Note*: this PR adds parameter descriptions for _all_ parameters that are already in the docstring (this commit does not add/remove any parameters to stay atomic). We believe this makes for cleaner PRs if this adds correct descriptions for all parameters and soon:

A follow-up PR (or several) after this is merged will deal with the parameters/arguments and:

- throw some out that users don't need and add some from sklearn that will be useful for texthero users.
- change some default values

### DIFF
--- a/texthero/representation.py
+++ b/texthero/representation.py
@@ -43,8 +43,10 @@ def representation_series_to_flat_series(
     ----------
     s : Sparse Pandas Series or Pandas Series
         The multiindexed Pandas Series to flatten.
+
     index : Pandas Index, optional, default to None
         The index the flattened Series should have.
+
     fill_missing_with : Any, default to np.nan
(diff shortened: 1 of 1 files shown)


jbesomi:
Very beautiful, really! 😍 

Only feedback: what if in the `Examples` we use `pd.pipe` as in the rest of the documentation (as we aim for unified documentation)?

henrifroese:
Thanks, just implemented using pipe. Also changed docstrings to load data in one line as you suggested in another PR (I believe the Visualization one)

jbesomi:
Amazing job! Thanks @henrifroese  and @mk2510 ! 💯

PR #121, opened 2020-07-24
Update function arguments (from sklearn) in Representation.

henrifroese:
- Remove arguments that are unnecessarily complex for most users
- Add arguments that are useful to many users.
- Revise default argument values (n_jobs=-1 where possible, copy_x=True where possible)

We believe that the arguments that are available to users now make the most sense for the community (as users with very complex needs will probably use sklearn anyway).

### DIFF
--- a/texthero/representation.py
+++ b/texthero/representation.py
@@ -107,6 +107,7 @@ def count(
     max_features: Optional[int] = None,
     min_df=1,
     max_df=1.0,
+    binary=False,
     return_feature_names=False,
 ) -> pd.Series:
     """
@@ -134,6 +135,9 @@ def count(
         frequency strictly higher than the given threshold.
         If float, the parameter represents a proportion of documents, integer
         absolute counts.
+
+    binary : bool, default=False
+        If True, all non zero counts are set to 1.
 
     return_features_names : Boolean, False by Default
         If True, return a tuple (*count_series*, *features_names*)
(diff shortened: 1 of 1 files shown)


henrifroese:
Just removed the `verbose` argument

texthero/representation.py
```diff
@@ -496,25 +504,17 @@ def nmf(s, n_components=2) -> pd.Series:
     :meth:`tfidf` to compute TF-IDF and :meth:`term_frequency` to compute term frequency
 
     """
-    nmf = NMF(n_components=n_components, init="random", random_state=0)
+    nmf = NMF(n_components=n_components, init="random", random_state=random_state,)
```
jbesomi, line 507: In contrast with `PCA`, we are not copying, is that expected?

jbesomi:
Super great! 🎉 🎉 

See the only comment below. By the way, should we keep the ":meth:`tfidf` to compute TF-IDF and :meth:`term_frequency` to compute term frequency"? Is this a bit out of scope maybe?

henrifroese:
> should we keep the ":meth:tfidf to compute TF-IDF and :meth:term_frequency to compute term frequency"? Is this a bit out of scope maybe?

Makes sense, it doesn't really belong in the "see also" section. Removed it.

jbesomi:
Great, thank you. Merged! 🎉 🍻

PR #124, opened 2020-07-26
Speed up preprocessing module

mk2510:
Looked at every function in the preprocessing module 😑 The only function we found, where it makes sense to speed up was the clean function 🧹  with the default pipeline. This is probably called with the default pipeline very often. The default pipeline loops through every function and we combined these into one pass in `_optimised_default_clean_single_cell(text: str)`, which is 30% faster than the default pipeline. Now when users call clean with the default pipeline, this function will be used. 
We understand, that this introduces code duality and more maintenance if bugs were fixed and changes to the default pipeline introduced, but we believe it makes sense, as the function is called so often. 🦝

### DIFF
--- a/texthero/preprocessing.py
+++ b/texthero/preprocessing.py
@@ -458,11 +458,67 @@ def clean(s: pd.Series, pipeline=None) -> pd.Series:
     """
 
     if not pipeline:
-        pipeline = get_default_pipeline()
+        return _optimised_default_clean(s)
 
     for f in pipeline:
         s = s.pipe(f)
     return s
+
+
+def _optimised_default_clean(s: pd.Series) -> pd.Series:
+    """
+    Applies the default clean pipeline in an optimised way to a series,
+    that is about 30% faster.
+
+    Default pipeline:
+     1. :meth:`texthero.preprocessing.fillna`
+     2. :meth:`texthero.preprocessing.lowercase`
+     3. :meth:`texthero.preprocessing.remove_digits`
+     4. :meth:`texthero.preprocessing.remove_punctuation`
+     5. :meth:`texthero.preprocessing.remove_diacritics`
+     6. :meth:`texthero.preprocessing.remove_stopwords`
+     7. :meth:`texthero.preprocessing.remove_whitespace`
+    """
+    return s.apply(_optimised_default_clean_single_cell)
+
+
+def _optimised_default_clean_single_cell(text: str) -> str:
+    """
+    Applies the default clean pipeline to one cell.
+
+    Default pipeline:
+     1. :meth:`texthero.preprocessing.fillna`
+     2. :meth:`texthero.preprocessing.lowercase`
+     3. :meth:`texthero.preprocessing.remove_digits`
+     4. :meth:`texthero.preprocessing.remove_punctuation`
+     5. :meth:`texthero.preprocessing.remove_diacritics`
+     6. :meth:`texthero.preprocessing.remove_stopwords`
+     7. :meth:`texthero.preprocessing.remove_whitespace`
+    """
+
+    # fillna
+    if pd.isna(text):
+        return ""
+
+    # lowercase
+    text = text.lower()
+
+    # remove digits and punctuation
+    pattern_digits_remove = r"\b\d+\b"
+    pattern_punctuation_remove = rf"([{string.punctuation}])+"
+    pattern_mixed_remove = pattern_digits_remove + "|" + pattern_punctuation_remove
+    text = re.sub(pattern_mixed_remove, "", text)
+
+    # remove diacritics
+    text = _remove_diacritics(text)
+
+    # remove stopwords
+    text = _replace_stopwords(text, _stopwords.DEFAULT, "")
+
+    # remove whitespace
+    text = " ".join(re.sub("\xa0", " ", text).split())
+
+    return text
 
 
 def has_content(s: pd.Series) -> pd.Series:


jbesomi:
Nice, thank you! 🎉 

Review:

1. What about `get_default_pipeline`? We can either keep it (but why?), remove it or rename `_optimised_default_clean` -> `get_default_pipeline`?
1. We need to extensively unit-test it as this will be one of the most used function
1. ... what if we have global constants regex?
1. What if we update the docstring of `clean` explaining the new changes?

mk2510:
> Review:
1.What about get_default_pipeline? We can either keep it (but why?), remove it or rename _optimised_default_clean -> get_default_pipeline?
2. We need to extensively unit-test it as this will be one of the most used function
3. what if we have global constants regex?
4.What if we update the docstring of clean explaining the new changes?

I have just updated the files to include those changes. 😅

texthero/preprocessing.py
```diff
@@ -17,6 +17,34 @@
 
 from typing import List, Callable
 
+# REGEX pattern constants
+PATTERN_REMOVE_DIGITS_BLOCK = r"\b\d+\b"
```
jbesomi, line 21: Probably, the words `REMOVE`, `REPLACE` and `PATTERN_` are not necessary:

PATTERN_REMOVE_DIGITS_BLOCK -> DIGITS_BLOCK
PATERN_REMOVE_CURLY_BRACKETS -> CURLY_BRACKETS
...

texthero/preprocessing.py
```diff
@@ -17,6 +17,34 @@
...
+PATERN_REMOVE_CURLY_BRACKETS = r"\{[^{}]*\}"
+PATTERN_REMOVE_SQUARE_BRACKETS = r"\[[^\[\]]*\]"
+PATTERN_REMOVE_ANGLE_BRACKETS = r"<[^<>]*>"
+PATTERN_REMOVE_HTML_TAG = r"""(?x)                    # Turn on free-spacing
+                            <[^>]+>                             # Remove <html> tags
+                            | &([a-z0-9]+|\#[0-9]{1,6}|\#x[0-9a-f]{1,6}); # Remove &nbsp;
+                            """
+
+
+def GET_PATTERN_TOKENIZATION(punct: str) -> str:
```
jbesomi, line 38: Global functions are always lowercased
"Returns the standart tokenisation pattern": not particularly meaningful
"Returns" -> "Return"

mk2510: ah thanks 🥇 I was unsure about the python style guide. But now it is a private function

mk2510:
I just have implemented those changes. Thx for the comments on python style guides 👍 
Also fixed some bigger merge problems in the preprocessing module

texthero/preprocessing.py
```diff
@@ -457,13 +472,67 @@ def clean(s: pd.Series, pipeline=None) -> pd.Series:
...
+
+    # fillna
+    if pd.isna(text):
+        return ""
+
+    # lowercase
+    text = text.lower()
+
+    # remove digits and punctuation
+    pattern_mixed_remove = DIGITS_BLOCK + "|" + PUNCTUATION
```
jbesomi, line 521: Is there a more elegant way of chaining regex? What are the best practices there?

texthero/preprocessing.py
```diff
@@ -17,6 +17,44 @@
...
+            """
+
+URLS = r"http\S+"
+TAGS = r"@[a-zA-Z0-9]+"
+HASHTAGS = r"#[a-zA-Z0-9_]+"
+
+
+def _get_pattern_for_tokenisation(punct: str) -> str:
+    """
+    Return the standart tokenisation pattern
```
jbesomi, line 44: word `standart` does not mean anything for most of the users ...
we should come up with a better explanation; i.e mentioning both what does this function does and why, which problem it solves

tests/test_preprocessing.py
```diff
@@ -114,6 +114,104 @@ def test_pipeline_stopwords(self):
         pipeline = [preprocessing.lowercase, preprocessing.remove_stopwords]
         self.assertEqual(preprocessing.clean(s, pipeline=pipeline), s_true)
 
+    """
+    Test clean
+    """
+
+    def _get_default_clean_pipeline(self):
```
jbesomi, line 121: Not sure we really need so many tests for this part ...?

jbesomi:
The new `clean` solution makes use of `apply`. But, in some cases (for instance with remove punctuation) we might not need to use apply, rather prefer `s.str.replace(REGEX)`, no?

I'm OK in improving the clean function but first I think we need to have a better understanding of which solution is actually the fastest, i.e we need to benchmark a bit. 

It would be great if you can do some benchmark and prepare some reports/insights, maybe as a single Jupiter notebook that you can attach here.

A possible benchmark pipeline might be:

1. apply(re.sub) vs str.replace
1. your version vs old version
1. go even further and try to parallelize more (with flashtext #87 for instance?)

We can also split this PR in two:

1. Add REGEX constants
1. Benchmark clean + add clean + add unit tests

tests/test_preprocessing.py
mk2510: I think those test will help us to cover all different sections of the pipeline individually, so if something gets changed, we know, which part is broken

texthero/preprocessing.py
mk2510: There is the re library to do this kind of stuff a bit more elegant, but I read, that str.replace gives us a performance improvement. As in my opinion, it is also quite clear, what it does, I think, we can keep it that way

texthero/preprocessing.py
mk2510: I have updated the comment. I hope it will be clearer now 🐰

mk2510:
>apply(re.sub) vs str.replace
your version vs old version

@jbesomi I have created here a notebook https://colab.research.google.com/drive/1HeVzomOS2F962qxTxKu_W__e1HaIfvZq?usp=sharing which presents the time differences. 

> go even further and try to parallelize more (with flashtext #87 for instance?)

as far, as I understood does the linked pull bring fasttext to our library. In my experience fasttext will be faster with huge regex pattern and not with the small ones we still have. So I think, this integration might be more part of the other issue 🏎️ 

@henrifroese and I will start with the multithreading task today, which will give general improvements. This pull request should just improve the functions by more efficient code 🤓

mk2510:
> apply(re.sub) vs str.replace

we need to do the re.sub as str.replace does not take any regex expressions when performed on a string.

PR #101, opened 2020-07-17
replace `tokenize_with_phrases` with `phrases` and added tests

cedricconol:
This PR will replace `tokenize_with_phrases` with `phrases`. I added unit tests as well for `phrases`.

This is the result of running `./tests.sh`:
```
..............................................................................................................................................................
----------------------------------------------------------------------
Ran 158 tests in 9.798s

OK
```

This is the result of running `./format.sh`:
```
All done! ✨ 🍰 ✨
6 files left unchanged.
All done! ✨ 🍰 ✨
6 files left unchanged.
```

PR #125, opened 2020-07-27
Add _check_tokenized function

AlfredWGA:
- Add a `_check_tokenized` function, avoid writing checking codes for each API.
- Add VS Code project setting folder `.vscode` to `.gitignore`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -184,3 +184,6 @@ dmypy.json
 # Cython debug symbols
 cython_debug/
 docs/source/api
+
+# Visual Studio Code
+.vscode/
--- a/texthero/representation.py
+++ b/texthero/representation.py
@@ -89,12 +89,18 @@ def representation_series_to_flat_series(
     return s
 
 
-# Warning message for not-tokenized inputs
-_not_tokenized_warning_message = (
-    "It seems like the given Pandas Series s is not tokenized. This function will"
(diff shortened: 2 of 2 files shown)


jbesomi:
Hey @AlfredWGA, thank you! 🎉 🎉 

Review:

1. That's a great idea, thanks! 👍 
1. What if we create a new private module `_helper` or something similar and move `_check_tokenized ` under there? As there will be other functions not in `representation` that will need to do the same check (see for instance #101)
1. As per #60, such Pandas Series `s` will be referenced as `TokenSeries`. What if we rename the method to `_is_tokenseries` or `_check_is_tokenseries` or something similar? In the future, then we will also have `is_representationseries` and so on
1. (minor comment) In the future, if that's possible, it would be great if you can keep the PR [self-contained](https://github.com/jbesomi/texthero/blob/master/CONTRIBUTING.md#best-practices) and avoid to do two different things. This is a [good article](https://opensource.zalando.com/blog/2017/10/small-pull-requests/) if you are interested in why. For this time, this is perfectly fine as the _gitignore_ changes are minimal

Regards,

jbesomi:
Note also the relation with #69

AlfredWGA:
Indeed #69 is a great solution for all these checking problem. I'll wait until that PR is completed and see what I can do further.

henrifroese:
Closing this as #69 is merged with this functionality.

PR #126, opened 2020-07-27
Implement 3D support for scatterplot

henrifroese:
Additionally:
- clean up the function
- add `hover_name` as argument
- improve docstring
- don't show figure when `return_figure` is set to True

### DIFF
--- a/tests/test_visualization.py
+++ b/tests/test_visualization.py
@@ -19,6 +19,26 @@ def load_tests(loader, tests, ignore):
 
 class TestVisualization(PandasTestCase):
     """
+    Test scatterplot.
+    """
+
+    def test_scatterplot_dimension_too_high(self):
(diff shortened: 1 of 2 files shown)


henrifroese:
> What does it happen if the input pca Series has only 1 column? Should we just let the program crash or put a custom more informative error message? + unit test

It's now handled.

> What does it happen if the PCA column has more than 3 columns? Do we just accept the first 3 or do we throw an informative error message? + unit test

Also handled.

> Unit-testreturn_figure when set to True

Done.

> Performance: we can obtain (x,y,z) with a single apply method instead of 2/3

I actually found no way to do this with `apply`, but what I now implemented with `np.stack` should be as fast as possible (as far as I can see).

> Parameters in the docstring and argument orders are different

Fixed.

jbesomi:
Thank you Henri and Max! 🎉

Issue #43, opened 2020-07-08
Support "Pandas Series Representation"


jbesomi:
This is one of the most interesting (future) aspects of Texthero: the ability to represent any text-dataset with ease, even very large dataset.

**Motivation**
One of the big limitations of the current version of the Texthero is that the output of the `tfidf` function or whatever other "representation" function is not particularly interpretable. The user do not even know which tf-idf weight is associated with which word/token.

The solution is to return a Multiindex Pandas Series where the first level represent the document and the second document represents the word. See this example below:

```
>>> import texthero as hero
>>> import pandas as pd
>>> s = pd.Series(["I am GROOT", "Flame on"])
>>> s = hero.tokenize(s)
>>> hero.tfidf(s)
document  word 
0        GROOT    0.577350
          I        0.577350
          am       0.577350
1        Flame    0.707107
          on       0.707107
dtype: Sparse[float64, nan]
```

The advantage of this approach is that:
1. The result is much more interpretable
2. The result is a **Sparse** Pandas Series! This is very good as often the output of TF-IDF (especially when max_features=None) is a very large and very sparse matrix. 

The drawback is that this Pandas Series cannot be appended directly into the Pandas Dataframe.

We refer to this MultiIndex series where the first level is the document and the second level is the term as 
  "Pandas Series Representation" (a better name is welcomed!)

**Texthero 2.0**

Starting from Texthero 2.0 all? "representation" functions will return such Pandas Representation Series. The `pca`/`nmf` function will accept as input a Pandas Representation and will (probably) return a flat representation as it does not make sense anymore to have a second level called "pca-component-1".

**From Pandas Representation Series to Pandas Series**

A function `to_flat_series` or something similar will transform the Pandas Representation Series into a (flatten) Pandas Series (as the actual output of `tfidf`). This will permit to append the Series into the initial df.

**From Pandas Representation Series to a document-term matrix**

Just by calling `.stack()` on the Pandas Representation Series it will be possible to convert it to a Pandas DataFrame where rows are the documents and every column is a term. Nice, right? We will need to explain clearly how to deal with MultiIndex (basics are not particularly hard)

**Interested in helping out?**
Most of the code has already been written. If you are interested in helping out for this **important** changes leave a comment. We will be glad to have you onboard!

**Your opinion**
Your opinion matter; let us know your thoughts!

jbesomi:
Useful code:


```

"""
Helper
"""


def _check_is_valid_representation(s) -> bool:
    """
[...]

PR #130, opened 2020-07-30
Implement & test handling of NaNs in preprocessing with handle_nans decorator

henrifroese:
- Decorator used everywhere it's necessary in preprocessing.py with appropriate fill values
- test_nan.py added with test's for the preprocessing module

### DIFF
--- a/tests/tests_nan.py
+++ b/tests/tests_nan.py
@@ -1,0 +1,103 @@
+"""
+Test whether all functions can handle NaNs.
+"""
+
+
+import pandas as pd
+import numpy as np
+from texthero import nlp, visualization, preprocessing, representation
+
+from . import PandasTestCase
+import unittest
+import string
+from parameterized import parameterized
+
+# Define valid inputs for different functions.
+s_nan = pd.Series(["test1", np.NaN, "test2", pd.NA])
+s_numeric_and_nan_lists = pd.Series([[5.0, 5.0], [6.0, 6.0], np.nan])
+
+"""
+Define all test cases. Every test case is a list
+of [name of test case, function to test, tuple of valid input for the function].
+First argument of valid input has to be the Pandas Series we
+want to test with at least one np.nan. If this is different for a function, a separate
+test case has to implemented in the class below.
+The tests will be run by AbstractNaNTest below through the @parameterized
+decorator.
+The names will be expanded automatically, so e.g. "named_entities"
+creates test cases test_ignores_nan_named_entities.
+"""
+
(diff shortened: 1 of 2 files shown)


jbesomi:
Thank you for the PR Henri!

I believe we are not dealing with _nan_ in the correct "Pandas" way. 

The rule should be:
 If input Pandas Series has some _nan_ at indexes X, output Pandas Series should have the same _nan_ at indexes X. 

**Preprocessing**:
 This can be "easily" reached in all preprocessing functions, probably at the exception of `phrases`. This PR is not necessary then.

Example: `remove_diacritics` yield

```
>>> s = pd.Series(["héllo", pd.Na])
>>> hero.remove_diacritics(s)
0. hello
1. (empty space)
```

Where the Pandas-user expects:

```
0. hello
1. pd.Na
```

**Representation**:

Here we need to deal with such NA values. The problems arises when we pass NA values to scikit learn functions, right?
The first question is:
 1. Is there a way we can pass NA to scikit learn function? If yes we don't even need `handle_nan`, rather just `test_na`

If that's not possible, your initial idea was to use a decorator as it might be useful. It keeps the code simple and does two things:
  1. Check and warns if `s` has NA
  1. Fill NA

Personally, I'm not a fan of the decorator in this case as the processes are a bit hidden. Instead, we can simply have a function: `s=_handle_nan(s)`. We might want to call it more explicitly `warn_and_fill_nans` or something similar. We need the same number of lines but we gain in clearness as we know exactly at which line the changes are made, we have a better idea of what `handle_nan` means and also we get some extra information regarding the input and output types.



My proposal is:

1. With `preprocessing` function: if na return na`
1. With `representation` function:
   1. Check if we cannot pass nan directly to scikitlearn
   1. If not we `fillna()`, compute and then either fill([], np.nan) or use your previous solution with the copy (we will need to evaluate the faster one)
1. Add `test_na` to verify the previously mentioned "rule" is respected

As you already spent quite a large amount of time, if you wish I can take care of this. 

P.s I'm sorry this is taking that long. It's surely my fault for not having explained this task properly as well as not having seen the issues initially.

jbesomi:
Update the description of issue #86
Also, this task is not our first priority yet, probably #43 is ...

mk2510:
Hi @jbesomi I think we will leave it till Monday and have a quick talk about it then 🚀

jbesomi:
Henri and Max, can you please just confirm you want me to take over this?

henrifroese:
Yes that would be great 👌

PR #134, opened 2020-07-31
Add normalization to representation module.

henrifroese:
- adds function `normalize` that takes Series and norm (supports `l1, l2, max` norms through `sklearn.preprocessing.normalize`)
- works on both representation series and flat / "normal" series (the sklearn normalize function can already handle both :clap: )
- after #132 is merged, will add more tests for normalize (doctests already cover the most important things)

### DIFF
--- a/texthero/representation.py
+++ b/texthero/representation.py
@@ -960,7 +960,7 @@ def normalize(s: pd.Series, norm="l2") -> pd.Series:
             s = s.astype("Sparse")
             s_coo_matrix = s.sparse.to_coo()[0]
 
-        s_for_vectorization = s_coo_matrix.todense()
+        s_for_vectorization = s_coo_matrix
 
     else:
         s_for_vectorization = list(s)
--- a/texthero/representation.py
+++ b/texthero/representation.py
@@ -10,6 +10,7 @@ from sklearn.manifold import TSNE
 from sklearn.decomposition import PCA, NMF
 from sklearn.cluster import KMeans, DBSCAN, MeanShift
 from sklearn.metrics.pairwise import cosine_similarity
+from sklearn.preprocessing import normalize as sklearn_normalize
 from scipy.sparse import coo_matrix
 
 from typing import Optional, Union, Any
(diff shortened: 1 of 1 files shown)


jbesomi:
😍 It's very beautiful how we can deal with the PandasRepresentation / sparse matrix, isn't it? Congrats for the clean code 🥇 

Unit-test after #132: yess! 

-> It would be great to add a new "pipeline" test file or something similar to test for more complex pipelines. For instance, it would be great to check that `tokenize` + `tfidf` + `norm` return the same output as `TFidfvectorizer` (clearly we can tune some default settings: lowercase=False, tokenize=..., remove_punctuation=no, ...). 

Remark: as we want to discourage the use of the flatten Pandas Series, are we sure we don't want to force the input to be a RepresentationSeries? Personally, I would prefer so as I don't see any advantage of passing/using a flatten Pandas Series if not than the RepresentationSeries cannot be stored in a Pandas Dataframe directly (as it has MultiIndex -> we will need to clearly mention this in the explanation and explain why we have the `flatten`).

To recap the goal of the `flatten` function:

1. (for now) to provide (half-)backward compatibility
1. (future) to let the user store a (previous) PandasRepresentation into a DataFrame
1. Nothing else, opinions?

henrifroese:
I fully agree with your goals for the `flatten` function, "2." seems like the most important one for users.

> It would be great to add a new "pipeline" test file or something similar to test for more complex pipelines

Yes, that probably makes sense later on; though I think it's not that high of a priority at the moment if the functions are well-covered by unit tests independently.

> as we want to discourage the use of the flatten Pandas Series, are we sure we don't want to force the input to be a RepresentationSeries? Personally, I would prefer so as I don't see any advantage of passing/using a flatten Pandas Series if not than the RepresentationSeries cannot be stored in a Pandas Dataframe directly

I'd keep normalization support for the flattened series as users might want to normalize other stuff that's not in representation series form but rather dense in their `df`, and the support for flat series is really easy. We might want to remove it later on but I think in the current state it still makes sense.

jbesomi:
> I'd keep normalization support for the flattened series as users might want to normalize other stuff that's not in representation series form but rather dense in their df, and the support for flat series is really easy. We might want to remove it later on but I think in the current state it still makes sense.

If we allow a `VectorSeries` as input it wouldn't be possible from a back-compatibility point of view to remove it. In the other hand, if we see that many users actually work with `VectorSeries` for some reason yet unknown we can always add this support. 

For keeping you on the discussion, in principle, we should entirely avoid using the `VectorSeries` as this is very inefficient. I cannot find it yet, but there is a rather infamous Stackoverflow answer that says clearly that Pandas Developers should avoid working with lists on Pandas Series. This is very inefficient from a memory and performance point of view. DataFrames should be used instead. 

In the perfect scenario, therefore, we would like to remove completely the `VectorSeries`. Sometimes this is very practical (for example the output of `pca`) and that's why we keep it. But just to deal with a very short and dense list. In Texthero's contest, normalization is done on a large and sparse matrix and that's why a `RepresentationSeries` as input is more natural.

Also, preferably every function should receive as input one single Series type. Again to keep things simple.

If you really believe we should normalize `VectorSeries`, then I would suggest adding a function `to_repr` or something like that that is the inverse of `flatten`. Again, I believe it's better not to add this for the reasons stated before but I'm open to opinions.

Let me know your thoughts, just trying to figure out what's will be the best for our community!

henrifroese:
Yes, I agree; we can leave it out to show users that they _shouldn't_ rely on vector series as performance gets really bad, and as you said:

> In the other hand, if we see that many users actually work with VectorSeries for some reason yet unknown we can always add this support.

texthero/representation.py
```diff
@@ -894,3 +895,74 @@ def meanshift(
...
+
+def normalize(s: pd.Series, norm="l2") -> pd.Series:
+    """
+    Normalize every cell in a Pandas Series.
+
+    Input has to be a Representation Series.
+
+    Parameters
+    ----------
+    s: Pandas Series
```
jbesomi, line 912: s: Pandas Series, RepresentationSeries (this is not that important yet but will become after we merge #69)

The sentence "Input has to be a Representation Series." is probably not strictly necessary and for consistency, we should either add it in every function or remove it

jbesomi:
Everything looks good (just see the minor review details), will merge it now! Thanks!! 🎉 🎉

Issue #94, opened 2020-07-15
Documentation Next steps: checklist


jbesomi:
This document attempt at giving an order of importance of tasks related to improving the documentation of the website.

- [ ] Add the "Edit on Github" button in every Texthero web-page
- [ ]

---
This document is a WIP

mk2510:
@jbesomi, as we now have implemented the API-checklist, we will now focus more on the documentation part. There we (@henrifroese and I) will start writing, the visualisation, representation, nlp tutorial and also a developer documentation :octocat: 
We will link the pull request later on here. 🥇

jbesomi:
Great; perfect!

PR #135, opened 2020-08-04
Tutorial on the different Series Types

henrifroese:
Only a first draft at the moment; critique welcome.

Focus on / extended explanation of Representation Series

Question: Is this the correct location for the file (`website/docs/`)?

### DIFF
--- a/website/docs/different_series_types_tutorial.md
+++ b/website/docs/different_series_types_tutorial.md
@@ -1,0 +1,86 @@
+<h1 align="center">Pandas Series Types in Texthero</h2>
+
+In Texthero, we're always working with Pandas Series and Pandas Dataframes to hold a (possibly very large) collection of documents. To make things easier and more intuitive, we differentiate between 4 different types of Series, depending on the cell content. For example, the functions in preprocessing.py usually take as input a Series where every cell is a string, and return as output a Series where every cell is a string.
+
+These are the implemented types:
+
+1. TextSeries: Every cell is a text, i.e. a string. For example,
+`pd.Series(["test", "test"])` is a valid TextSeries.
+
+2. TokenSeries: Every cell is a list of words/tokens, i.e. a list
+of strings. For example, `pd.Series([["test"], ["token2", "token3"]])` is a valid TokenSeries.
+
+3. VectorSeries: Every cell is a vector representing text, i.e.
+a list of floats. For example, `pd.Series([[1.0, 2.0], [3.0]])` is a valid VectorSeries.
+
+4. RepresentationSeries: Series is multiindexed with level one
+being the document, level two being the individual features and their values.
+For example,
+`pd.Series([1, 2, 3], index=pd.MultiIndex.from_tuples([("doc1", "word1"), ("doc1", "word2"), ("doc2", "word1")]))`
+is a valid RepresentationSeries.
+
+Now, if you see a function in the documentation that looks like this:
+```
+def tfidf(s: TokenSeries) -> RepresentationSeries
+```
+
+then you know that the function takes a Pandas Series
+whose cells are lists of strings (tokens) and will
+return a Pandas Series whose cells are lists of floats. And this function:
+```
+def pca(s: Union[VectorSeries, RepresentationSeries) -> VectorSeries
+```
+can handle both _VectorSeries_ and _RepresentationSeries_ as input and always returns a _VectorSeries_.
+
+
+
+<h2 align="center">Representation Series</h2>
+
+As you can see, the `RepresentationSeries` type is a little more complex than the others. Let's have a closer look to see what it is and why, where and how it is used!
+
+<h3 align="left">What is it?</h2>
+
+A _RepresentationSeries_ is multiindexed with level one
+being the document, and level two being the individual features and their values. It could look like this:
+
+```python
+>>> import texthero as hero
+>>> import pandas as pd
+>>> s = pd.Series(["Sentence one one", "Sentence two"])
+>>> s.pipe(hero.tokenize).pipe(hero.count)  # first tokenize Series, then calculate word count
+document  word    
+0         Sentence    1
+          one         2
+1         Sentence    1
+          two         1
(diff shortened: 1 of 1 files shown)


jbesomi:
> Only a first draft at the moment; critique welcome (you can have a look at it e.g. by pasting everything in here.

You can directly add it to the website so that we can see it by clicking "view deployment".

Also, "view file" do the job: [view file](https://github.com/jbesomi/texthero/blob/5fb3551fb7a4f6eb6d1d13d5e5d2c35a32d7616d/website/docs/different_series_types_tutorial.md).

> Question: Is this the correct location for the file (website/docs/)?

See #94 for a better contextualization. The idea is to have this file under the "Getting Started" tab.

We might want to call this file: `getting-started-heroseries.md`


Review:

1. Very good start 💪 
1. The style should be aligned somehow with the [getting-started](https://texthero.org/docs/getting-started) document. I would start a bit gently by showing some examples and explain a bit deeper.
1. Before actually writing the content, we can just discuss the structure, the different headings, and sub-headings and the content of each sub-heading, without writing it?


Structure (just a draft)

#HeroSeries

One line intro to give an idea on the subject

## Overview

Explain the why
Show the 4 

## TextSeries

...

## TokenSeries

...

## RepresentationSeries

...

## VectorSeries

- mention better to avoid, used for `pca` as practical



Open question:

We can have either this page just after the "Getting Started" page or at the end.

Version 1:

1. Getting Started
1. Preprocessing
1. Representation
1. NLP
1. Visualization
1. HeroSeries

Version 2:

1. Getting Started
1. HeroSeries
1. Preprocessing
1. Representation
1. NLP
1. Visualization

We should consider which solution works best, particularly taking into consideration the content of the other pages.

The advantage of Version 1 is that the user understand immediately what HeroSeries are and stand for.

Advantage of Version 2 is that the user got to see the whole pipeline in the details and he should be more aware of the difference between the Series. On the other hand, it's also true that the "1. Getting Started" already provides a general introduction to the pipeline.

**Long story short, we need to have clear in mind the aim of this documentation page as well as how it will interface with the rest of the documentation. The idea again here is to provide a simple and unified view of Texthero.** 

I believe we can aim toward a results similar to the docusaurus documentation: [doc](https://docusaurus.io/docs/en/installation).

Extra:
- We need to explain clearly that these are just Pandas Series, nothing else

henrifroese:
> Advantage of Version 2 is that the user got to see the whole pipeline in the details and he should be more aware of the difference between the Series. On the other hand, it's also true that the "1. Getting Started" already provides a general introduction to the pipeline.

We agree with this. Getting started provides an overview, then we have a deep-dive for the users really interested in the different modules through Preprocessing, Representation, NLP, Visualization; and if they still want to know more, they get to the HeroSeries stuff. I think many users will read the getting started, and then later on when googling they will land on `texthero.org` and then see the `TextSeries, RepresentationSeries, ...` in the docstrings. This will lead them to the tutorial too as we'll link it in the docstrings!

Also, we just improved the content / explained in more detail, and changed to your proposed structure.

website/docs/getting-started-herotypes.md
```diff
@@ -0,0 +1,148 @@
...
+
+<h3 align="left">DataFrame</h3>
+
+In Natural Language Processing, we are often working with matrices that contain information about our dataset. For example, the output of the functions `tfidf`, `count`, and `term_frequency` is a [Document Term Matrix](https://en.wikipedia.org/wiki/Document-term_matrix), i.e. a matrix where each row is one document and each column is one term / word.
+
+We use a Pandas DataFrame for this for two reasons:
+1. It looks nice.
+2. It can be sparse.
+
+The second reason is worth explaining in more detail: In e.g. a big Document Term Matrix, we might have 10,000 different terms, so 10,000 columns in our DataFrame. Additionally, most documents will only contain a small subset of all the terms. Thus, in each row, there will be lots of zeros in our matrix. This is why we use a [sparse matrix](https://en.wikipedia.org/wiki/Sparse_matrix): A sparse matrix only stores the non-zero fields. And Pandas DataFrames support sparse data, so Texthero users fully profit from the sparseness!
```
mk2510, line 123: Document-term and every where else as well

website/docs/getting-started-herotypes.md
```diff
@@ -0,0 +1,148 @@
...
+
+This is a massive advantage when dealing with *big data*: In a _sparse DataFrame_, we only store the data that's relevant to save lots and lots of time and space!
```
mk2510, line 125: maybe: big data sets, as with big data the user might think of something else

website/docs/getting-started-herotypes.md
```diff
@@ -0,0 +1,148 @@
...
+3            "Text in the fourth document"
+4                                    ...
+
+```
+
+ Consequently, in Texthero's _preprocessing_ module, the functions usually take as input a Series where every cell is a string, and return as output a Series where every cell is a string. We will call this kind of Series _TextSeries_, so users know immediately what kind of Series the functions can work on. For example, you might see a function
+ ```python
+remove_punctuation(s: TextSeries) -> TextSeries
+ ```
+in the documentation. You then know that this function takes as input a _TextSeries_ and returns as output a _TextSeries_, so it can be used on a DataFrame or Series in the preprocessing phase of your work, where each document is one string.
```
mk2510, line 23: I don't think it can be used on a dataframe as well, just on the textcolumn, which will then be interpreted as a series. So this sentence might be a bit confusing

website/docs/getting-started-herotypes.md
henrifroese: will do :paintbrush:

website/docs/getting-started-herotypes.md
henrifroese: done

website/docs/getting-started-herotypes.md
henrifroese: yes, you're right - will change :running_man:

PR #136, opened 2020-08-05
Fix docstring line length in NLP

cedricconol:
Limited `nlp` docstring line length to 75 characters following numpy's docstring guide. I have to manually format the docstring since `black` does not auto-format it. If this is approved, will do the same for other modules.


jbesomi:
Beautiful, thank you! I encourage you also to read the docstring and eventually improve it/make it more unified

cedricconol:
Sure! I'll open another PR for that.

PR #138, opened 2020-08-05
Use custom Pandas Series Types in Preprocessing

henrifroese:
- use the types from `_helper.py`
- use decorator `InputSeries` where it makes sense


jbesomi:
Thank you, very nice!

In here or in a separate PR we should also update the docstring "Parameter". 

Before:

```
    Parameters
    ----------
    s : Pandas Series
    ...
```

After

```
    Parameters
    ----------
    s : RepresentationSeries or s : Pandas Series, RepresentationSeries
    ...
```

The sentence "Input has to be a Representation Series." such as the one under `normalize` might not be necessary anymore. We can either remove it everywhere or add it everywhere ... if you don't want to check that as it's a bit annoying and long you can also open a new issue, someone else might want to help with that as we can label it "good first issue". 

I'm fine if we deal with that in a separate PR, just let me know which solution is better for you

henrifroese:
Done :heavy_check_mark:

texthero/preprocessing.py
```diff
@@ -14,16 +14,17 @@
...
 from texthero import stopwords as _stopwords
+from texthero._helper import TokenSeries, TextSeries, InputSeries
 
-from typing import List, Callable
+from typing import List, Callable, Union
 
 # Ignore gensim annoying warnings
 import warnings
 
 warnings.filterwarnings(action="ignore", category=UserWarning, module="gensim")
```
jbesomi, line 26: Missing `@InputSeries(TextSeries)`?

jbesomi:
Henri, what if we create a separate file `_types.py` that deal with that? This is more than general helper functions, isn't it? If we do it right now, we can avoid doing it later when we would need to change more ...

texthero/preprocessing.py
henrifroese: I believe you mean it's missing for `fillna`. If yes, just added it

henrifroese:
Just did this:

- move HeroSeries typing stuff from `_helper.py` to new `_types.py`
- move the tests to new `test_types.py` file accordingly
- add type check to `fillna`

jbesomi:
Good job! Thank you Henri and Max! 🎉 🎉

PR #139, opened 2020-08-05
Use custom Pandas Series Types in NLP and Visualization

henrifroese:
- use the types from `_helper.py`
- use decorator `InputSeries` where it makes sense

Comment: will implement this in `representation.py` after all the Representation Series stuff is merged


jbesomi:
Same review as #138

henrifroese:
I believe what you describe in #138 is not necessary/present anywhere in `nlp.py` or `visualization.py`

jbesomi:
My bad, sorry! Thank you! 🎉

Issue #137, opened 2020-08-05
Unit-test/Travis CI: cleaner log


jbesomi:
The current Travis CI log isn't particularly beautiful, [example](https://travis-ci.com/github/jbesomi/texthero/jobs/368704825).

This task consists of cleaning the log output of travis CI by mainly improving the unit-tests files.

Also, sometimes the Travis CI does not fail even if some errors are raised, see example below at line 576 for instance (`raise IOError(Errors.E050.format(name=name))). This should also investigate further. In this case, we may want the job to fail rather than pass?

PR #146, opened 2020-08-09
Support for Flair Embeddings: hero.embed(s, flair_embedding) function

henrifroese:
- new file `embeddings.py`
- function `hero.embed(s: TokenSeries, flair_embedding: flair.embeddings.DocumentEmbeddings) -> VectorSeries`
- tests for everything in new `test_embeddings.py`
- new dependency `flair>=0.5.1.`

The new function adds full support for [Flair Document Embeddings](https://github.com/flairNLP/flair/blob/master/resources/docs/TUTORIAL_5_DOCUMENT_EMBEDDINGS.md). See `embeddings.py` docstrings for further explainations.

### DIFF
--- a/texthero/embeddings.py
+++ b/texthero/embeddings.py
@@ -1,0 +1,117 @@
+"""
+embeddings.py
+-------------
+
+Embed documents and work with the embeddings to find similar documents,
+find topics, ... .
+
+There are many different ways to transform text data into vectors to gain
+insights from them. The resulting vectors are called _embeddings_.
+A _word embedding_ assigns a vector to each word, a _document embedding_
+(sometimes called thought _thought vector_)
+assigns a vector to each document.
+
+One way to get embeddings is to use a function such as tfidf,
+count, or term_frequency that creates vectors depending
+on which words occur how frequently.
+
+Another option is to use embeddings that try to directly capture
+the semantic relationship between words or sentences. For example,
+the vector of 'biochemistry' minus the vector of 'chemistry' is
+close to the vector of 'biology'.
+
+In Texthero, both options are supported. The second option is
+implemented through the `Flair library <https://github.com/flairNLP/flair>`_
+
+
+TODO :
+- think about:
+    - Do we (want to) support word embeddings? How? Why?
+    - Any opportunity for sparseness or alternatives to s.apply?
+"""
+
+import pandas as pd
+import numpy as np
+
+import flair
+from flair.data import Sentence, Token
+
+from typing import List, Union
+
+from texthero._types import InputSeries, TokenSeries, VectorSeries
+
+
+"""
+Helper functions.
+"""
+
+
+def _texthero_init_for_flair_sentence(self, already_tokenized_text: List[str]):
+    """
+    To use flair embeddings, flair needs as input a
+    'flair.Sentence' object. Creating such an object
+    only works from strings in flair. However, we want
+    our embeddings to work on TokenSeries, so we
+    overwrite the 'flair Sentence' __init__ method with
+    this method to create Sentence objects from already
+    tokenized text.
+    """
+
+    super(Sentence, self).__init__()
+
+    self.tokens: List[Token] = []
+    self._embeddings: Dict = {}
+    self.language_code: str = None
+    self.tokenized = None
+
+    # already tokenized -> simply add the tokens
+    for token in already_tokenized_text:
+        self.add_token(token)
+
+
+# Overwrite flair Sentence __init__ method to handle already tokenized text
+Sentence.__init__ = _texthero_init_for_flair_sentence
+
+
+"""
+Support for flair embeddings.
+"""
+
+
+@InputSeries(TokenSeries)
+def embed(
+    s: TokenSeries, flair_embedding: Union[WordEmbeddings, DocumentEmbeddings]
+) -> VectorSeries:
+    """
+    TODO
+    Examples
+    --------
+    >>> import texthero as hero
+    >>> import pandas as pd
+    >>> from flair.embeddings import TransformerDocumentEmbeddings
+    >>> embedding = TransformerDocumentEmbeddings('bert-base-uncased')
+    >>> s = pd.Series(["Text of doc 1", "Text of doc 2"]).pipe(hero.tokenize)
+    >>> hero.embed(s, embedding)  # doctest: +SKIP
+    0    [-0.6618074, -0.20467158, -0.05876905, -0.3482...
+    1    [-0.5505255, -0.21915795, -0.0913163, -0.26856...
+    dtype: object
+
+    """
+
+    def _embed_and_return_embedding(x):
+        # flair embeddings need a 'flair Sentence' object as input.
+        x = Sentence(x)
+        # Calculate the embedding; flair writes it to x.embedding
+        flair_embedding.embed(x)
+        # Return it as numpy array.
+        return x.embedding.detach().numpy()
+
+    if isinstance(flair_embedding, DocumentEmbeddings):
+        s = s.apply(lambda x: _embed_and_return_embedding(x))
+    else:
+        raise ValueError(
+            "Unknown embedding type. Texthero only works with"
+            " flair DocumentEmbeddings."
+        )
+
+    return s
--- a/texthero/embeddings.py
+++ b/texthero/embeddings.py
@@ -22,19 +22,16 @@ close to the vector of 'biology'.
 
 In Texthero, both options are supported. The second option is
 implemented through the `Flair library <https://github.com/flairNLP/flair>`_
-
-
-TODO :
-- think about:
-    - Do we (want to) support word embeddings? How? Why?
-    - Any opportunity for sparseness or alternatives to s.apply?
 """
 
 import pandas as pd
 import numpy as np
 
+# We only `import flair` and
+# call everything from flair directly,
+# e.g. flair.data.Tokens, to not pollute our
+# namespace with similarly sounding names.
 import flair
-from flair.data import Sentence, Token
 
 from typing import List, Union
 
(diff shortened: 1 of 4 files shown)


henrifroese:
Build fails on Windows and errors on macOS, will look into it

jbesomi:
Nice and very beautiful! 🎉 

Why is `flair` not automatically installing `torch`? Shouldn't we specify `install_require=...torch...` in setup.cfg too instead of in the travis yaml file?

What about the discussion/questions there: #140. Q: `embed` is also a  representation function, shouldn't we return a `RepresentationSeries`? What's your view on that?

A distinction we might want to do is that in general the `embed`/`dim red` size is quite limited (100-300 dim size) and as such `VectorSeries` still make sense, whereas for `tfidf/count/term_freq` the size can be as large as the number of tokens and therefore the use of `RepresentationSeries`...

texthero/embeddings.py
```diff
@@ -0,0 +1,123 @@
...
+
+from texthero._types import InputSeries, TokenSeries, VectorSeries
+
+
+"""
+Helper functions.
+"""
+
+
+def _texthero_init_for_flair_sentence(self, already_tokenized_text: List[str]):
```
jbesomi, line 46: Very beautiful solution!! 💯 💯 !!

texthero/embeddings.py
```diff
@@ -0,0 +1,123 @@
...
+flair.data.Sentence.__init__ = _texthero_init_for_flair_sentence
+
+
+"""
+Support for flair embeddings.
+"""
+
+
+@InputSeries(TokenSeries)
+def embed(
```
jbesomi, line 79: The first idea was to have this under `representation`, i.e to keep the number of modules fix to four ... opinion?

henrifroese: Our thinking was to move everything that _takes TokenSeries and produces a vector per text_, so `TokenSeries -> VectorSeries` and `TokenSeries -> RepresentationSeries` under one `embeddings.py` file, so we'd move `tfidf, count, term_frequency` there too after this is merged.

However, we could also put this into representation, probably just a question of how we see the trade-off _long files vs. many files_.

henrifroese:
> Why is flair not automatically installing torch? Shouldn't we specify install_require=...torch... in setup.cfg too instead of in the travis yaml file?

Flair tries to automatically install torch, but it fails on macOS and Windows. We could probably open a separate Issue/PR after this is merged to clean up the pipeline etc. , similar to #137 . You can e.g. see [here](https://pytorch.org/get-started/locally/) when selecting the Windows configuration that it won't just install with `pip install torch`

> What about the discussion/questions there: #140. Q: embed is also a representation function, shouldn't we return a RepresentationSeries? What's your view on that?

I'll answer on #140 soon, but generally: It doesn't really make sense to return a RepresentationSeries here as we don't have any sparseness, as you say:

> A distinction we might want to do is that in general the embed/dim red size is quite limited (100-300 dim size) and as such VectorSeries still make sense, whereas for tfidf/count/term_freq the size can be as large as the number of tokens and therefore the use of RepresentationSeries...

So overall, for embeddings (`embed, tfidf, count, term_frequency`) the first one gives a VectorSeries, the other three give a RepresentationSeries. We now have the big problem that `pca, nmf, tsne` only support RepresentationSeries. I see two options:

1. Re-implement VectorSeries input support for `pca, nmf, tsne` (we already had that but decided to throw it out in #140 )
2. Change `embed` such that it returns a RepresentationSeries. We'd not profit from the sparseness, of course, but it would also _not be more expensive that a VectorSeries_.

texthero/embeddings.py
jbesomi: I agree that we don't want very long files (but in my experience sometimes it's easier to find things in a long document with CTRL-F rather than search in multiple files and folders and understand a more complex structure). I also understand and share your same view on the trade-off.

What I would suggest is that we can split the representation file into two files but we hide this to users, i.e in the API all these functions will appear under `texthero.representation` (that's what I meant by saying to keep only 4 modules)

henrifroese: Yes makes sense. Maybe I should move this `embeddings.py` stuff to `representation` for the time being and we can do the re-org in a separate PR?

jbesomi:
If I well understood what you are saying, on a clean macOS/windows machines,`pip install texthero` won't work as `torch` will not installed correctly? What's your idea of how we can solve this?

What's a bit unclear is that I haven't found anything on the [flair](https://github.com/flairNLP/flair) documentation that explains how `torch` should be installed. It appears like`pip install flair` should do the job!

henrifroese:
Hmm I'm actually not sure (and also not sure how to try that out), I'll try again with some `.travis.yml` configurations in this PR

henrifroese:
We have now tried out lots of stuff in #149 .

- Linux works :heavy_check_mark: 
- Windows: installing `torch` just doesn't work with `pip install torch`; see [here](https://pytorch.org/get-started/locally/) that it does not work with Windows; `pip install flair` also won't work on Windows as long as `torch` is not already installed
- macOS: installing `pip install texthero or flair etc.` does usually work, the torch dependency can however produce errors when `--user` is not set as flag

We see the following options:

1. Remove `flair` as default dependency (would save the huge `torch` dependency and only affects the `embed` function) -> when users want to use `hero.embed`, check if `flair` is installed and tell them to install it if it isn't

2. Use the `.travis.yml` file that is currently used in this PR and tell macOS users in the `README` they might want to use `--user` if they have problems (most won't have any issues) and tell Windows users how they can ~buy a Mac~ install `torch` first

What do you think? :upside_down_face:

jbesomi:
1. Seems ok for now. Anyway, the users will have to understand how `flair` works, so it's "okay" not to have it as dependencies (at least for now). We will have to make this clear in the documentation (docstring/getting started/tutorial).

We still need to "fake" on Travis CI and install flair as you were doing (and torch eventually) otherwise, unit-tests will fail

> can ~buy a Mac~ install torch first

😂😂😂😂

henrifroese:
Great, we've now switched flair to a dev dependency and made the necessary changes in the code :octocat:

.travis.yml
```diff
@@ -3,25 +3,47 @@ jobs:
   include:
     - name: "Python 3.6.0 on Xenial Linux"
       python: 3.6
+      install: 
+      - pip3 install --upgrade pip  # all three OSes agree about 'pip3'
+      - pip3 install black
```
jbesomi, line 8: `black` is in the [dev dependencies](https://github.com/jbesomi/texthero/blob/master/setup.cfg), why do we need it there?

texthero/embeddings.py
```diff
@@ -0,0 +1,134 @@
...
+In Texthero, both options are supported. The second option is
+implemented through the `Flair library <https://github.com/flairNLP/flair>`_
+"""
+
+import pandas as pd
+import numpy as np
+
+# We only `import flair` and
+# call everything from flair directly,
+# e.g. flair.data.Tokens, to not pollute our
```
jbesomi, line 32: 👌

texthero/embeddings.py
jbesomi: Yes!

texthero/embeddings.py
```diff
@@ -0,0 +1,134 @@
...
+    s: TokenSeries, flair_embedding: flair.embeddings.DocumentEmbeddings
+) -> VectorSeries:
+    """
+    Generate a vector for each document using the given flair_embedding.
+
+    Given a tokenized Series and a
+    `Flair Document Embedding https://github.com/flairNLP/flair/blob/master/resources/docs/TUTORIAL_5_DOCUMENT_EMBEDDINGS.md`_,
+    return the document embedding for every
+    document in the tokenized Series.
+
```
jbesomi, line 90: We need to add the Parameters docstring section

texthero/embeddings.py
```diff
@@ -0,0 +1,134 @@
...
+    Examples
+    --------
+    >>> import texthero as hero
+    >>> import pandas as pd
+    >>> from flair.embeddings import TransformerDocumentEmbeddings
+    >>> embedding = TransformerDocumentEmbeddings("google/reformer-crime-and-punishment")
+    >>> s = pd.Series(["Text of doc 1", "Text of doc 2"]).pipe(hero.tokenize)
+    >>> hero.embed(s, embedding)  # doctest: +SKIP
```
jbesomi, line 98: `# doctest: +SKIP` necessary?

texthero/embeddings.py
```diff
@@ -0,0 +1,134 @@
...
+    `Flair Document Embedding https://github.com/flairNLP/flair/blob/master/resources/docs/TUTORIAL_5_DOCUMENT_EMBEDDINGS.md`_
+    """
+    try:
+        import flair
+    except:
+        raise ImportError(
+            "To use this function, you need to have"
+            " the flair library (https://github.com/flairNLP/flair)"
+            " installed and imported!"
+        )
```
jbesomi, line 114: Add a `TODO` comment: "Add a link to the Flair/Texthero article that explains how to flair with Texthero and how to install flair+torch" (Flair documentation lack clarification on this point ...)

texthero/embeddings.py
```diff
@@ -0,0 +1,134 @@
...
+
+    super(flair.data.Sentence, self).__init__()
+
+    self.tokens: List[flair.data.Token] = []
+    self._embeddings: Dict = {}
+    self.language_code: str = None
+    self.tokenized = None
+
+    # already tokenized -> simply add the tokens
+    for token in already_tokenized_text:
```
jbesomi, line 65: The `for` cycle is quite inefficient, can't we do better?

Would be great if we can transform the whole `TokenSeries` into a sort of `FlairTokenSeries` in a more efficient way ...

henrifroese:
TODO:
- clearer description that (and how) users have to install flair themselves
- longer explanation in the docstring

PR #147, opened 2020-08-09
Representation: `most_similar`, closes #45

### LINKED ISSUE #45: Add most_similar(s, distance_algorithm, vector)
Add `hero.most_similar(s, representation, distance_algorithm, threshold, vector)`

This task is very similar to #4.

Given a Pandas Series, the code returns a Pandas Series ordered by the distance from the vector.

**Open question** (same question as #4)

In case of `s` is a Text Series, `representation` is required.
In case of `s` is already represented (i.e is a Representation Series), `representation` is not required.

We can either accept that s has both structures or stick to only one (in this case probably the second solution is preferable).
Opinions on that?

henrifroese:
- add function `most_similar` in `representation.py`

Only a draft at the moment; no tests yet.

Interested in opinions concerning the "function design"! It currently has signature `hero.most_similar(s: Union[RepresentationSeries, VectorSeries], vector: List[float], max_number=None)` and works e.g. like this:
```
>>> import texthero as hero
>>> import pandas as pd
>>> s = pd.Series(["I like football", "Hey, watch out", "I like sports", "Cool stuff"])
>>> s_pca = s.pipe(hero.tokenize).pipe(hero.tfidf).pipe(hero.flatten).pipe(hero.pca)

>>> # want to find the two most similar to "I like football", which has index 0
>>> s_most_similar = hero.most_similar(s_pca, s_pca[0], max_number=2)
>>> s_most_similar
0                       [0.0]
2    [2.1073424255447017e-08]
dtype: object

>>> # get text of the most similar ones (of course, the text is most similar to itself)
>>> s[s_most_similar.index]
0    I like football
2      I like sports
dtype: object
```

So it essentially returns a Series with the same index as the input series, but sorted by similarity descending.

### DIFF
--- a/texthero/representation.py
+++ b/texthero/representation.py
@@ -10,10 +10,11 @@ from sklearn.manifold import TSNE
 from sklearn.decomposition import PCA, NMF
 from sklearn.cluster import KMeans, DBSCAN, MeanShift
 from sklearn.metrics.pairwise import cosine_similarity
+from sklearn.metrics import pairwise_distances
 from sklearn.preprocessing import normalize as sklearn_normalize
 from scipy.sparse import coo_matrix
 
-from typing import Optional, Union, Any
+from typing import Optional, Union, Any, List
 
 from texthero import preprocessing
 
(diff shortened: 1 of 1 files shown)


henrifroese:
Put on hold until Series Types are figured out.

henrifroese:
- maybe use cosine distance by default; add argument to choose function
- maybe change signature to `most_similar(s: TokenSeries(?), text, representation_function, metric)`
- maybe prepare tutorial for users instead of adding function to TextHero

PR #152, opened 2020-08-18
Speed up Series initialization

henrifroese:
switch to `list(x)` instead of `x.tolist()` for quick speedup.

To test the performance :racing_car: :

```
matrix = np.random.normal(size=(100000, 500))
# new
%timeit pd.Series(list(matrix))
# old
%timeit pd.Series(matrix.tolist())

```

### DIFF
--- a/texthero/representation.py
+++ b/texthero/representation.py
@@ -87,7 +87,7 @@ def flatten(
         # are present in the Sparse Representation Series
         # "reappear" correctly.
 
-    s = pd.Series(s.values.tolist(), index=s.index)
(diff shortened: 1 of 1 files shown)


jbesomi:
Great, thank you! 🎉

jbesomi:
```
38.1 ms ± 2.17 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
3.35 s ± 41.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

About 80x faster! Crazy. Do we know why?

PR #118, opened 2020-07-22
Update getting_started.md

mk2510:
*removed spelling mistakes from the getting started 

* added tokenization functions to the document and process

PR #117, opened 2020-07-22
Update README.md

mk2510:
* corrected some spelling mistakes

* added tokenize to the demo code snippets

* explanation of which functions take tokenized input

PR #156, opened 2020-08-21
Change representation_series to DataFrame

mk2510:
* all functions, which previously dealt with representation series now handle only the dataframe instead. 🚀 

* rm all functions like flatten, as they are not needed anymore

* adopted docstrings and tests

-> further stuff to do: 
  - add those examples into the tutorials, readme, getting started


jbesomi:
Looks great (even too hard to understand, at least quickly).

The main question (and sorry for the late review; will catch up faster now): why do we return a MultiIndex sparse DataFrame? Why not simply a (sparse) DataFrame? This should simplify things a bit and probably is the most natural type users expect (i.e we wrap the scipy sparse matrix on a DF)

texthero/representation.py
```diff
@@ -136,22 +56,18 @@ def count(
     min_df=1,
     max_df=1.0,
     binary=False,
-) -> pd.Series:
+) -> pd.DataFrame:
     """
     Represent a text-based Pandas Series using count.
 
-    Return a Document Representation Series with the
+    Return a Document Term DataFrame with the
```
jbesomi, line 63: The newline formatting is a bit bizarre there + the docstring is not crystal clear.

What about something like that?

"""
Return a count document-term DataFrame based on the given Pandas Series

Rows of the returned DataFrame represent document whereas columns are terms ...


"""

mk2510:
> Why not simply a (sparse) DataFrame? 

We introduced the Multicolumndataframe to better structure the returned data. In this way, the user can access the tfidf and gets a matrix data frame returned.
Image 03 09 20 at 21 20

So, when inserting the tfidf data frame into a big one, the user could use the following syntax:

```python
data = pd.read_csv("file.csv")
data["tfidf"] = hero.tfidf(data["clean"])
data["lda"] = hero.lda(data["clean"])
```
This way, the user can use his workflow, which he is used when using pandas, because in this example `data["tfidf"]` returns the tfidf-Matrix

jbesomi:
Hey Max, sure; I see what you mean!

> This way, the user can use his workflow, which he is used when using pandas,

Until now, users do not really make use of _columns_ MultiIndex ...

Also, users most probably expect that `tfidf` return a document-term `DataFrame`.

I would solve the problem the other way around; i.e for `tfidf`, `count`, ..., we return simple (sparse) DF and eventually in #158 we will first wrap the DataFrame into a MultiIndex DataFrame to allow the insertion into others DFs.

mk2510:
> Until now, users do not really make use of columns MultiIndex ...
I totally agree with you there. 🔥 Even, when it makes displaying them a little bit prettier, it adds way too much confusion to the user and other developers. We will fix this issue and change PR #158 as well. So, when the user inserts a MatrixDF into his existing one, we will convert that DF into a multicolumn one, so the user can have the typical pandas workflow even with matrixes. 🥳 🎉

texthero/representation.py
mk2510: changed the docstring into a clearer version

mk2510:
@jbesomi we now have finally implemented all changes to integrate the DataFrame into the representation functions. Also the PRs #157, #158, #117, #118, #135 were adopted to include all changes. 🔥 code and documentation wise we should be ready to introduce the DataFrame type and release a new version to pip. 🐙

jbesomi:
That's amazing! I will review and test everything on Monday morning! Hopefully Monday evening or Tuesday is merged 🎉🍻
Awesome job!

texthero/representation.py
```diff
@@ -701,14 +616,8 @@ def kmeans(
...
     >>> import pandas as pd
-    >>> s = pd.Series(["Football, Sports, Soccer", "music, violin, orchestra",
-    ...                "football, fun, sports", "music, fun, guitar"])
-    >>> s = (
-    ...     s.pipe(hero.clean)
-    ...      .pipe(hero.tokenize)
-    ...      .pipe(hero.term_frequency)
-    ...      .pipe(hero.flatten)
-    ... ) # TODO: when others get Representation Support: remove flatten
+    >>> s = pd.Series(["Football, Sports, Soccer", "music, violin, orchestra", "football, fun, sports", "music, fun, guitar"])
```
jbesomi, line 619: docstring larger than [max characters allowed ](https://github.com/jbesomi/texthero/blob/master/CONTRIBUTING.md)(75)

texthero/representation.py
```diff
@@ -442,9 +324,13 @@ def pca(s, n_components=2, random_state=None) -> pd.Series:
...
     represented to a matrix form.
 
+    PCA cannot directly handle sparse input, so when calling pca on a
+    DocumentTermDF, the input has to be expanded which can lead to
+    memory problems with big datasets.
+
     Parameters
     ----------
-    s : Pandas Series
+    s : Pandas Series (VectorSeries) or MultiIndex Sparse DataFrame (DocumentTermDF)
```
jbesomi, line 333: MultiIndex Sparse DataFrame (DocumentTermDF): DataFrame is enough (change that also in the other functions below)

texthero/representation.py
```diff
@@ -542,11 +438,18 @@ def nmf(s, n_components=2, random_state=None) -> pd.Series:
...
+        s_for_vectorization = s_coo.astype("float64")
+    else:
+        s_for_vectorization = list(s)
+
+    return pd.Series(list(nmf.fit_transform(s_for_vectorization)), index=s.index)
 
 
 def tsne(
-    s: pd.Series,
+    s: Union[pd.Series, pd.DataFrame],
```
jbesomi, line 452: Calling `s` a variable that accepts both `s`eries and Dataframe is confusing ... ideas for a better name?

jbesomi:
Overall, that's a super great job 🎉 , we just need to do some work of refinement.

Major comments:
- We shouldn't mention `MultiIndex` in the whole PR, right?
- `pca` and so on can work on document-term matrix but also in other kind of DataFrame kinds (as long as they have numeric-only values)
- `term_frequency`, `count` and `tfidf` should have the same docstring structure (similar first sentence, ...)
- Important: for the clustering function, we are not testing all their parameters. Shouldn't we do something similar to `test_vectorization_not_tokenized_yet_warning `?
- It's not ideal that we are skipping a lot of docstring tests ...

.travis.yml
```diff
@@ -20,7 +20,7 @@ jobs:
       env: PATH=/c/Python38:/c/Python38/Scripts:$PATH
 install: 
     - pip3 install --upgrade pip  # all three OSes agree about 'pip3'
-    - pip3 install black
+    - pip3 install black==19.10b0
```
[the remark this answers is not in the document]
mk2510: that is the latest black version, which does not destroy our doctests. The new black-version removes some whitespaces because of which the doctests fail. We are keeping track of it in Issue #171

texthero/representation.py
mk2510: I totally agree with you there. I changed the name to `input_matrix`, as we always get a matrix as input but in different representations, so instead of calling it similar to the datatype we called it `input_matrix` as we get a matrix input. 🇫🇯

texthero/representation.py
```diff
@@ -209,37 +124,26 @@ def count(
...
+    )
 
 
 def term_frequency(
     s: pd.Series, max_features: Optional[int] = None, min_df=1, max_df=1.0,
-) -> pd.Series:
+) -> pd.DataFrame:
     """
-    Represent a text-based Pandas Series using term frequency.
+    Return a count document-term DataFrame based on the given Pandas Series
```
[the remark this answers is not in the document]
mk2510: you are totally right there 🥇

texthero/representation.py
```diff
@@ -266,17 +170,14 @@ def term_frequency(
...
     >>> import pandas as pd
     >>> s = pd.Series(["Sentence one hey", "Sentence two"]).pipe(hero.tokenize)
-    >>> hero.term_frequency(s)
-    0  Sentence    0.2
-       hey         0.2
-       one         0.2
-    1  Sentence    0.2
-       two         0.2
-    dtype: Sparse[float64, nan]
+    >>> hero.term_frequency(s) # doctest: +SKIP
```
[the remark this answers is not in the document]
mk2510: we will try out the `# doctest: +NUMBER` option, as this should ignore less relevant floats. 🤞 we now hope it will work 🔥

texthero/representation.py
```diff
@@ -27,94 +27,14 @@
...
-        # )
-
-    if s.index.nlevels != 2:
-        return False
-        # raise ValueError(
-        #     f"The input Pandas Series should be a Representation Pandas Series and should have a MultiIndex, where the first level represent the document and the second one the words/token. The given Pandas Series has {s.index.nlevels} number of levels instead of 2."
-        # )
-
-    return True
+    return isinstance(df, pd.DataFrame) and not isinstance(df.columns, pd.MultiIndex)
```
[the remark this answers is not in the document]
mk2510: we use this function for example in the function `normalize` or in the functions which take an input_matrix and when converting this matrix to an object, which `sklearn.PCA` can handle.

tests/test_representation.py
```diff
@@ -168,69 +206,65 @@ def test_vectorization_arguments_to_sklearn(self, name, test_function, *args):
...
-        s_true = pd.Series(
-            [[3.0, "FILLED", np.nan], ["FILLED", 4.0, "FILLED"]],
-            index=["doc0", "doc1"],
-        )
+        if name == "normalize":
+            # testing this below separately
+            return
+
+        if name == "kmeans":
+            result_s = test_function(s_documenttermDF, random_state=42, n_clusters=2)
```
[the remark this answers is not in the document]
mk2510: that does definitely sounds like a reasonable idea. But I would propose to open an issue and new PR, which refractors our tests. This way, we introduce those architectural changes separately and not messed up with other important topics, like DataFrame support :octocat:

tests/test_representation.py
```diff
@@ -50,39 +50,114 @@ def _tfidf(term, corpus, document_index):
...
+        representation.tfidf,
+        pd.DataFrame([2, 1], index=s_tokenized_output_index, columns=["Test"],).astype(
+            "Sparse[float64, nan]"
+        ),
+    ],
+]
+
+
+s_vector_series = pd.Series([[1.0, 0.0], [0.0, 0.0]], index=[5, 7])
+s_documenttermDF = pd.DataFrame(
```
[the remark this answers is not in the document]
mk2510: done ✅

tests/test_representation.py
```diff
@@ -50,39 +50,114 @@ def _tfidf(term, corpus, document_index):
...
+        "tfidf",
+        representation.tfidf,
+        pd.DataFrame([2, 1], index=s_tokenized_output_index, columns=["Test"],).astype(
+            "Sparse[float64, nan]"
+        ),
+    ],
+]
+
+
+s_vector_series = pd.Series([[1.0, 0.0], [0.0, 0.0]], index=[5, 7])
```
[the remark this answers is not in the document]
mk2510: done ✔️

tests/test_representation.py
```diff
@@ -50,39 +50,114 @@ def _tfidf(term, corpus, document_index):
...
-s_tokenized_output_index = pd.MultiIndex.from_tuples(
-    [(0, "!"), (0, "TEST"), (0, "Test"), (1, "."), (1, "?"), (1, "Test")],
-)
-
-s_tokenized_output_noncontinuous_index = pd.MultiIndex.from_tuples(
-    [(5, "!"), (5, "TEST"), (5, "Test"), (7, "."), (7, "?"), (7, "Test")],
-)
-
-s_tokenized_output_min_df_index = pd.MultiIndex.from_tuples([(0, "Test"), (1, "Test")],)
+s_tokenized_output_index = pd.Index([0, 1])
```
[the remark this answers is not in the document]
mk2510: that sounds like a very reasonable rule of thumb. But I would again suggest, to implement those in a new PR for all variables in tests, as otherwise, the changes will get very confusing. 🥴
However, I changed the newly introduced and obvious variables.

tests/test_representation.py
```diff
@@ -168,69 +206,65 @@ def test_vectorization_arguments_to_sklearn(self, name, test_function, *args):
...
-            [[3.0, "FILLED", np.nan], ["FILLED", 4.0, "FILLED"]],
-            index=["doc0", "doc1"],
-        )
+        if name == "normalize":
+            # testing this below separately
+            return
+
+        if name == "kmeans":
+            result_s = test_function(s_documenttermDF, random_state=42, n_clusters=2)
+        elif name == "dbscan" or name == "meanshift" or name == "normalize":
```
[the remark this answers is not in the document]
mk2510: ☑️

texthero/representation.py
mk2510: done ✅

mk2510:
@jbesomi regarding your points

> We shouldn't mention MultiIndex in the whole PR, right?

I totally agree with you there. I removed it from where I could find it, as it is just a normal dataframe 🤞 

> pca and so on can work on document-term matrix but also in other kind of DataFrame kinds (as long as they have numeric-only values) 

that's also right. I totally forgot to change here the name of the checking function as well. Only did it in the PR for the DataFrame types. 🤦 

> term_frequency, count and tfidf should have the same docstring structure (similar first sentence, ...)

With count and term frequency I totally agree with you there. However I think the short explanation of the tfidf formula in the docstring can be quite handy for the users, but in count and term frequency I think explaining it, would be too much

> Important: for the clustering function, we are not testing all their parameters. Shouldn't we do something similar to test_vectorization_not_tokenized_yet_warning 

It might be worth thinking about it in the issue of test environment refactoring. 🤔 but it would not suit this PR

> It's not ideal that we are skipping a lot of docstring tests ...

it is definatly not ideal, but right now I can't see a way on how to integrate float doctest in an elegant way. 🙈

jbesomi:
As far as I know, we will not have to deal anymore with "RepresentationSeries" as there are no functions that return such object. `tfidf` and company returns a DataFrame that we can convert into a VectorSeries. Is that right?

Then, for instance, `normalize` does not need to handle the `RepresentationSeries` case, rather `VectorSeries` and `DataFrame`. It seems to me that `_check_is_valid_DataFrame` is not useful as-it-is, rather, we can move the code inside the `normalize` function to detect if `input_matrix` is a VectorSeries (need to check is valid) or simply a DataFrame.

There are still many part of the code that mention "DocumentTermDF", this might not be necessary, right? i.e `pca` can be applied to any DataFrame, not only document-term ...

+ plus you need to fix minor issues, such as docstring longer than 75 characters ... (i.e line 917), `pca` and `nmf` not having the same summary sentence (one is missing "on the given input.") [I haven't checked the others]

> ☝️ @mk2510 for the next time, make sure to set "ready for review" when you double-checked the whole PR and you are sure on all changes. It makes spare time to both you and me :) 👍

mk2510:
However we use `_check_is_valid_DataFrame` in multiple different functions, like `pca`, `mnf`, etc. It is necessary to convert the given data into the right function input format. This works differently for VectorSeries and for DataFrames. Hence we use this function quite often. If only used in normalize, I totally agree with you, that it would be better to move it inside the function. 

> There are still many part of the code that mention "DocumentTermDF", this might not be necessary, right? i.e pca can be applied to any DataFrame, not only document-term ...

I totally miss those 🤦  but now all unnecessary Document Term mentions should be gone.

>  pca and nmf not having the same summary sentence (one is missing "on the given input.") [I haven't checked the others]

Those summary sentences should now be the same and the lengths also under 76 everywhere 🤞 

> As far as I know, we will not have to deal anymore with "RepresentationSeries" as there are no functions that return such object. tfidf and company returns a DataFrame that we can convert into a VectorSeries. Is that right?
 

That is absolutly right. The representation Series will be removed in the next PR, where we worked on the hero types.

texthero/representation.py
```diff
@@ -27,101 +28,22 @@
...
-              Word3    NaN
-    doc1      Word2    4.0
-    dtype: float64
-    >>> hero.flatten(s, fill_missing_with=0.0)
-    document
-    doc0    [3.0, 0.0, nan]
-    doc1    [0.0, 4.0, 0.0]
-    dtype: object
-
+def _check_is_valid_DataFrame(df: Union[pd.DataFrame, pd.Series]) -> bool:
```
henrifroese, line 31: There isn't really a reason why users would have a multiindex in their DataFrame. Thus, this function can be replaced by just `isinstance(df, pd.DataFrame)`

texthero/representation.py
```diff
@@ -136,42 +58,38 @@ def count(
...
+) -> pd.DataFrame:
     """
     Represent a text-based Pandas Series using count.
 
-    Return a Document Representation Series with the
-    number of occurences of a document's words for every
-    document.
+    Rows of the returned DataFrame represent document whereas 
+    columns are terms. The value in the cell document-term is
+    the number of the term inthis document. The output is sparse.
```
henrifroese, line 67: typo

texthero/representation.py
```diff
@@ -180,17 +98,18 @@ def count(
...
+    >>> hero.count(s) # doctest: +SKIP        
+      Sentence one two
+    0        1   1   0
+    1        1   0   1
+   
     See Also
     --------
-    Document Representation Series: TODO add tutorial link
+
+    DataFrame: TODO add tutorial link
```
henrifroese, line 112: Would remove the "DataFrame"

texthero/representation.py
```diff
@@ -209,74 +128,62 @@ def count(
...
+    ...                                                     )
+    >>> hero.term_frequency(s) # doctest: +SKIP              
+            Sentence  hey  one  two
+    0            0.2  0.2  0.2  0.0
+    1            0.2  0.0  0.0  0.2
 
     See Also
     --------
-    Document Representation Series: TODO add tutorial link
+    DataFrame: TODO add tutorial link
```
henrifroese, line 186: as above

texthero/representation.py
```diff
@@ -357,33 +257,31 @@ def tfidf(s: pd.Series, max_features=None, min_df=1, max_df=1.0,) -> pd.Series:
...
+        Bye        Hi      Test
+    0   1.0  1.405465  0.000000
+    1   2.0  0.000000  1.405465
 
     See Also
     --------
     `TF-IDF on Wikipedia <https://en.wikipedia.org/wiki/Tf-idf>`_
 
-    Document Representation Series: TODO add tutorial link
+    DataFrame: TODO add tutorial link
```
henrifroese, line 284: same as above

henrifroese (commented):
Just went through everything once more. Will fix the very small stuff I found. It's then ready to merge in my opinion

texthero/representation.py
```diff
@@ -136,42 +48,38 @@ def count(
...
 
-    Use :meth:`hero.representation.flatten` on the output to get
-    a standard Pandas Series with the document vectors
-    in every cell.
-
     Parameters
     ----------
     s : Pandas Series (tokenized)
 
     max_features : int, optional, default to None.
```
jbesomi, line 67: replace it with `default=None` (to be consistent with the rest), change also in the other functions

texthero/representation.py
```diff
@@ -180,17 +88,18 @@ def count(
...
     >>> import pandas as pd
-    >>> s = pd.Series(["Sentence one", "Sentence two"]).pipe(hero.tokenize)
-    >>> hero.count(s)
-    0  Sentence    1
-       one         1
-    1  Sentence    1
-       two         1
-    dtype: Sparse[int64, 0]
-
+    >>> s = pd.Series(["Sentence one", "Sentence two"]).pipe(
```
jbesomi, line 91: Looks a bit ugly and `>>> s = pd.Series(["Sentence one", "Sentence two"]).pipe(hero.tokenize)` is 71 characters (less than 75, equal if we consider the initial tab of 4 characters). Let's keep in in one line :)

Same discussion for the other functions

texthero/representation.py
```diff
@@ -180,17 +88,18 @@ def count(
...
+    ...                                    hero.tokenize
+    ...                                                )
+    >>> hero.count(s) # doctest: +SKIP
```
jbesomi, line 94: remove # doctest: +SKIP

texthero/representation.py
```diff
@@ -136,42 +48,38 @@ def count(
...
     binary=False,
-) -> pd.Series:
+) -> pd.DataFrame:
     """
     Represent a text-based Pandas Series using count.
 
-    Return a Document Representation Series with the
-    number of occurences of a document's words for every
-    document.
+    Rows of the returned DataFrame represent document whereas
```
jbesomi, line 55: "represent document**s**"

change also in `term_frequency` + `tfidf`

texthero/representation.py
```diff
@@ -326,26 +223,19 @@ def tfidf(s: pd.Series, max_features=None, min_df=1, max_df=1.0,) -> pd.Series:
...
-    formula described above.
+    Different from the `sklearn-implementation of tfidf
+    <https://scikit-learn.org/stable/modules/generated/sklearn.feature_
+    extraction.text.TfidfVectorizer.html>`, this function does *not* 
+    normalize the output in any way, so the result is exactly what you 
+    get applying the formula described above.
 
-    Return a Document Representation Series with the
-    tfidf of every word in the document.
+    Return a Document Term DataFrame with the
```
jbesomi, line 232: Drop these two lines as they are redundant (already written above)

jbesomi:
Looks almost perfect 😍 

I just noticed how we don't have a strict rule for how we define the default value in the docstring. I believe we can stick to :

```
max_features : int, optional, default=None
```

Can you please make sure in all functions we write it the same way?

texthero/representation.py
henrifroese: Will rather do this in a separate PR across all files (see comment below)

henrifroese:
> Can you please make sure in all functions we write it the same way?

Yes, I'll do that and add the "British/American English" and "number of default components" to `CONTRIBUTING.md` and change it in all files later.

texthero/representation.py
henrifroese: :heavy_check_mark:

texthero/representation.py
henrifroese: :heavy_check_mark:

texthero/representation.py
henrifroese: I'll try that but I'm pretty sure this will sadly fail on MacOS in travis as it prints the DF differently

texthero/representation.py
henrifroese: done for all where it's possible

henrifroese:
Just incorporated the suggested changes from the review :zap:

henrifroese:
As we can see, the DF doctest fails in macOS, so I'll skip it again

texthero/representation.py
jbesomi: OK. As a general comment: we need to find a general stable solution; this problem is taking us too much time. The initial idea of having doctests was to just make sure that the output of the Example section corresponded to the output of the Texthero functions; that's it. Say we change a hero function, doctests help us make sure we update also the docstring Example section accordingly.

I believe that it make sense somehow to make sure that the output of the Examples corresponds to the output of the functions, yet, we don't wanto to get crazy if the outputs from mac and windows change just a bit or if we forget a whitespace somewhere.

What I propose is to stop testing the doctest as we are doing now and write our own wrapper to test that. By looking at the unittest and using for instance [inspect](https://docs.python.org/3/library/inspect.html) I think this will be feasible. We might want also to write a "tech" article on the blog page or something like that as this might be useful for others too. 

What are your opinions and ideas?

jbesomi:
Ok, see [here](https://github.com/jbesomi/texthero/pull/156#discussion_r487394297)

texthero/representation.py
henrifroese: I agree that the doctests are annoying. I'll think about this some more - it makes sense to not use the default module as it gets really annoying now. I'll look into alternatives like e.g. a decorator

texthero/representation.py
jbesomi: OK!

jbesomi:
Let's go! 🎉 🎉  Good job.

PR #157, opened 2020-08-22
HeroTypes in Representation; DataFrame in _types

henrifroese:
- switch `DocumentTermDF` in for `RepresentationSeries` in `_types.py`
- add functionality for decorator `@InputSeries` to handle several allowed input types
- Add typing decorator/hints to `representation.py`
- add tests for DocumentTermDF type in `test_types.py`

*NOTE*: only so many commits/lines as this builds on #156


mk2510:
we have now also adopted the type checks. The given DataFrame should not be a multicolumn one, but a "normal" Sparse DataFrame. We decided to not call it SparseDataFrame, as in the future the DataFrame may not have to be sparse. :octocat:

texthero/representation.py
```diff
@@ -14,6 +14,13 @@
...
 from typing import Optional, Union, Any
+from texthero._types import (
+    TextSeries,
+    TokenSeries,
+    VectorSeries,
+    DataFrame,
+    InputSeries,
+)
 
 from texthero import preprocessing
```
AlfredWGA, line 25: Can we remove this import as mentioned in #128?

henrifroese: I would not remove this here as we're trying our best to keep the pull requests atomic. I think it's best to do this in a separate PR

AlfredWGA: Sure thing! looking forward to your merge.

jbesomi:
Nice!

Just realizing that this PR solve some of the issues of #180

I will review the code details once #156 is merged.

Overall questions: why do we need the `DocumentTermDF` type? That's just a DataFrame with numeric-only values, right? As discussed, I would not have a specific type for this, at least for now.

mk2510:
> Overall questions: why do we need the DocumentTermDF type? That's just a DataFrame with numeric-only values, right? As discussed, I would not have a specific type for this, at least for now.

We changed it to only DataFrame, as it also could represent a different constellation than DocumentTermDF. But when we remove this type from our pandas types, we have an Incomplete TypeCheck. For now the DataFrame typcheck is not as much worth as the Series Types, but it completes the typing of every input.

henrifroese:
We could of course just use the actual `pd.DataFrame` type in our type checks. Arguments:

- Pro: easier for us; not confusing to users
- Con: our documentation will then not link to a hero series article with a nice docstring explaining that this is to represent a matrix etc (as the _types module is not only useful for typechecks but also for documentation).

henrifroese:
We have now re-merged #156 in this and solved the last remaining issues we saw. Ready to review/merge now from our side :3rd_place_medal:

texthero/_types.py
```diff
@@ -43,8 +42,8 @@ def tfidf(s: TokenSeries) -> RepresentationSeries:
 - TextSeries: cells are text (i.e. strings), e.g. "Test"
 - TokenSeries: cells are lists of tokens (i.e. lists of strings), e.g. ["word1", "word2"]
 - VectorSeries: cells are vector representations of text, e.g. [0.25, 0.75]
-- RepresentationSeries: Series is multiindexed with level one
-being the document, level two being the individual features and their values
+- DataFrame: a Pandas DataFrame representing a matrix. Every cell value
+    is one entry in the matrix.
 
 The classes are lightweight subclasses of pd.Series and serve 2 purposes:
```
jbesomi, line 48: Overall, the introduction is super great but not anymore 100% accurate.

I would add a separate paragraph about the new changes this PR make:

"Sometimes texthero functions also accept (standard) Pandas `DataFrame`s (in general, when they accept `VectorSeries`, is there any exception?). For this reason HeroSeries subclass also DataFrame ... "

Plus, would drop the "DataFrame: a Pandas ..." item from "These are the implemented types" as is not really an implementation.

texthero/_types.py
```diff
@@ -81,68 +82,70 @@ class HeroSeries(pd.Series):
     3. VectorSeries: Every cell is a vector representing text, i.e.
     a list of floats. For example, `pd.Series([[1.0, 2.0], [3.0]])` is a valid VectorSeries.
 
-    4. RepresentationSeries: Series is multiindexed with level one
-    being the document, level two being the individual features and their values.
+    4. DataFrame: A Pandas DataFrame representing a matrix.
```
jbesomi, line 85: The same discussion as before; just mention we sub-class also from DataFrame to allow this type as input for specific functions

texthero/_types.py
```diff
@@ -185,45 +190,50 @@ def is_numeric(x):
...
+                return False, error_string
         except KeyError:  # Only NaNs in Series -> same warning applies
-            raise TypeError(error_string)
+            return False, error_string
+
+        return True, ""
 
 
-class RepresentationSeries(HeroSeries):
+class DataFrame(HeroTypes):
```
jbesomi, line 205: Is this class really necessary?

tests/test_representation.py
```diff
@@ -232,7 +232,7 @@ def test_dim_reduction_and_clustering_with_vector_series_input(
         )
 
     @parameterized.expand(test_cases_dim_reduction_and_clustering)
-    def test_dim_reduction_and_clustering_with_documenttermDF_input(
+    def test_dim_reduction_and_clustering_with_DataFrame_input(
```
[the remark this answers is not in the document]
henrifroese: will do

texthero/_types.py
henrifroese: will change this

texthero/_types.py
henrifroese: will change

texthero/_types.py
henrifroese: Yes, it makes checking the types much easier for us through `@InputSeries` and gives users a good error message / docstring. We _could_ also overwrite the Pandas DataFrame class to add the function `check_type`, but that's at least as much code and annoying for other developers. This implementation makes it easy to use with the decorator.

jbesomi: Ok; I see. Sure, let's keep as it is, maybe just adding a sentence explaining why (the same answer you gave it to me there, it might be helpful for other developers too!)

henrifroese: Yes, I added an explanation to the `_types` file docstring :kangaroo: :ok_hand:

henrifroese:
From our side, this is now ready to be re-reviewed / merged :crossed_fingers: :baby_bottle:

jbesomi:
Thanks! Merged 🎉 🎉

henrifroese:
Great! 🎉🎉🎉. We will now go through all PRs in the merge checklist that are branched from this and change them accordingly! 💡👍

PR #158, opened 2020-08-22
Add pandas __setitem__ support for DocumentTermDF

mk2510:
As discussed, this PR makes small changes to `pd.DataFrame.__setitem__` to allow users to do e.g. `df["tfidf"] = hero.tfidf(df["text"])`. See the extensive documentation we added in `_helper.py`. We also add tests for this 🕵️ 🕵️‍♂️ 🕵️‍♀️  

*NOTE*: only so many commits/lines as this builds on #156


mk2510:
@jbesomi we now have also adapted the set_item method to handle the changes in PR #156 when the set item method receives a data frame, which has more than one column it will convert the columns into a pandas multi-index with the original column names on the lower level and the `set key` on the upper level like the user would naturally expect how pandas work. 🐼 
so to demonstrate the functioning, when a DataFrame with multiple columns is inserted into an existing one: 
```python
>>> df1 = pd.DataFrame(["Text 1", "Text 2"], columns=["Test"])
>>> df2 = pd.DataFrame([[3, 5], [8, 4]], columns=["term 1", "term 2"],)
>>> df1["count"] = df2
>>> df1
     Test  count       
          term 1 term 2
0  Text 1      3      5
1  Text 2      8      4
```
when just a single column DataFrame is inserted, it will behave, as it used to:
```python
>>> df1 = pd.DataFrame(["Text 1", "Text 2"], columns=["Test"])
>>> df2 = pd.DataFrame([3, 5], columns=["count"])
>>> df1["here"] = df2
>>> df1
     Test  here
0  Text 1     3
1  Text 2     5
```

jbesomi:
Magic. 🥇 

I will review this once #156 is merged so that we can test a bit around with the output DataFrame(s).

Does your solution handle correctly the insertion of sparse Pandas DF?

For you to know, my only concern is that I don't fully understand why the Pandas API does not allow for such insertion.

mk2510:
> For you to know, my only concern is that I don't fully understand why the Pandas API does not allow for such insertion.

There is no reason, that the pandas API does not support it. [This](https://github.com/pandas-dev/pandas/issues/35727) issue we opened at pandas about it. Now our usecase is much easier, with a single column level.

> Does your solution handle correctly the insertion of sparse Pandas DF?

It works so far. As soon, as the other two are merged, I will move this PR to ready to review

henrifroese (commented):
We just went through everything again and added additional tests for sparseness. We believe this is ready to be reviewed/merged now :octocat: :octopus:

texthero/_helper.py
```diff
@@ -71,3 +71,166 @@ def wrapper(*args, **kwargs):
...
+
+Additionally, we support this for pd.concat in a similar way; again, see the
+implementation below for details.
+
+Advantages / Why does this work?
+
+    - we don't destroy any pandas functionality as currently calling
+      `__setitem__` with a Multiindexed value is just not possible, so
+      our changes to Pandas do not break any Pandas functionality for
+      the users. We're only _expanding_ the functinoality
```
jbesomi, line 146: functinoality spelling mistake

texthero/_helper.py
```diff
@@ -71,3 +71,166 @@ def wrapper(*args, **kwargs):
...
+```
+>>> hero.count(df["text"])
+ count                  
+      1  2  3 Text doc of
+0     1  0  0    1   1  1
+1     0  1  0    1   1  1
+2     0  0  1    1   1  1
+```
+
+That's a DataFrame. Great! Of course, users can
```
jbesomi, line 105: we can be a bit more succint here (~That's a DataFrame. Great! Of course,~)

jbesomi (commented):
Hmm, less complex than expected 🧐
Looks great!

What if we add this for now as an extension that users can test? We can insert this into a separate file `texthero.beta.helper` or similar and explain in a blog article how to "activate that" (from texthero.beta import helper ...) and what it allows us to do.

mk2510:
I also discussed this issue with @henrifroese. we decided that we will not merge it, as discussed in the meeting. In our view when hiding this feature in a beta version, most users will probably ignore this. As this is performance-wise not optimized for huge DataFrames we decided that the tradeoff is not worth it and we will close this PR

jbesomi:
Ok. I'm sorry that we couldn't merge it, as it would have been a great feature for users, yet, it would have been probably worse to introduce something that would have made the user experience less pleasant. The fact that assigning a sparse DataFrame takes so long because of the large number of columns is what makes me think twice before making it part of the master branch. Let's hope that with Pandas v.2 we will be able to do something similar in a more efficient way 🙏🏻

PR #163, opened 2020-08-25
Implementation of Topic modeling and visualisation of these

mk2510:


henrifroese:
As discussed, we'll do some more work on this

mk2510:
@jbesomi We implemented the suggested changes. Big functions `visualize_topics` and `top_words_per_document` are now split up into multiple smaller ones, so the user of the library now has to write the pipeline of functions by himself. The new functions include:

*topic_matrices*
```python
def topic_matrices(s_document_term: pd.DataFrame, s_document_topic: pd.Series):
```
 Get a DocumentTopic Matrix and a TopicTerm Matrix (both as Dataframes) from a DocumentTerm Matrix and a DocumentTopic Matrix.  Both these matrices (the first one relating documents to terms and the second one relating documents to topics) are used to generate a DocumentTopic Matrix (relating documents to topics) and a TopicTerm Matrix (relating topics to terms).

*relevant_words_per_topic*
```python
def relevant_words_per_topic(s_document_term, s_document_topic_distribution, s_topic_term_distribution, n_words=10, return_figure=False):
```
Use [LDAvis](https://github.com/bmabey/pyLDAvis)  to find the most relevant words for each topic. This function uses the three given relations (documents->terms, documents->topics, topics->terms) to find and return the most relevant words for each topic. The pyLDAvis library is used to find relevant words.

We left the two functions _visualize_topics_ and _relevant_words_per_document_ inside the library as we think, that there will be a lot of users, especially ML specialist, who just want to have a short glimpse of their data and not understand the whole algorithm. However, in the docstring we mentioned, that those functions are just a pipeline wrapper like we did it with the clean function in preprocessing. 

For `relevant_words_per_document` the pipeline could look like:
```python
>>> # New Series where every document is its own cluster.
>>> s_cluster = pd.Series(
...    np.arange(len(s_document_term)), index=s_document_term.index, dtype="category")  
>>> s_document_topic, s_topic_term = hero.topic_matrices(s_document_term, s_cluster) 
>>> s_document_topic_distribution = hero.normalize(s_document_topic, norm="l1")
>>> s_topic_term_distribution = hero.normalize(s_topic_term, norm="l1")  
>>> relevant_words_per_topic(
...  s_document_term,
...  s_document_topic_distribution,
...  s_topic_term_distribution)  # doctest: +SKIP
```

and for `visualize_topics` a suggested hero-pipeline would look like this:
```python
>>> import pyLDAvis  # doctest: +SKIP
>>> s_document_topic, s_topic_term = hero.topic_matrices(s_document_term, s_document_topic) 
>>> s_document_topic_distribution = hero.normalize(s_document_topic, norm="l1") 
>>> s_topic_term_distribution = hero.normalize(s_topic_term, norm="l1") 
>>> figure = hero.relevant_words_per_topic(s_document_term, s_document_topic_distribution, s_topic_term_distribution, return_figure=True) 
>>> pyLDAvis.show(figure)
```

jbesomi:
Looks great! 🎉 🎉  👍 

Can you please provide a short Google Cloab notebook that shows a working pipeline with all the different functions integrated?

Do these functions works also on a large dataset? Up until which size?

mk2510:
@jbesomi we created a [short](https://colab.research.google.com/drive/15U0XW_dfr05mB3ATIktH7P4s3Mz0Z3qg?usp=sharing) notebook, where we display the functionality of those two pipelines. 🐰 We think those two will be the main use cases of the implemented functions. The third 🥉 use case, finding relevant words per topic and including them in a data frame is just a version of the second algorithm, but with the documents clustered with a clustering algorithm like kmeans or assigned to a topic with LSA/LDA. 

When those functions will be ready to merge, we will prepare an exhaustive tutorial for the user to introduce them to the Topic Modeling 💯

texthero/representation.py
```diff
@@ -947,22 +893,730 @@ def meanshift(
...
+
+    result = pd.Series(
+        list(truncatedSVD.fit_transform(s_for_vectorization)), index=s.index
+    )
+
+    return result
+
+
+def lda(
+    s: Union[pd.Series, pd.DataFrame],
```
jbesomi, line 973: Same comment as #156

texthero/representation.py
```diff
@@ -947,22 +893,730 @@ def meanshift(
...
+    n_components=10,
+    max_iter=10,
+    random_state=None,
+    n_jobs=-1,
+) -> pd.Series:
+    """
+    Performs Latent Dirichlet Allocation on the given pandas series.
```
jbesomi, line 980: Not only pandas series + we should have written as "Pandas Series" for consistency with the rest of the documentation

texthero/representation.py
```diff
@@ -947,22 +893,730 @@ def meanshift(
...
+
+    Latent Dirichlet Allocation (LDA) is a topic modeling algorithm 
+    based on Dirichlet distribution. In natural language processing
+    LDA is often used to categorise documents into diffenrent topics
```
jbesomi, line 984: Texthero's docstring uses the English style rather than the British style (TODO we should add it under CONTRIBUTING.md). Replace categorise with categorize

texthero/representation.py
```diff
@@ -947,22 +893,730 @@ def meanshift(
...
+    and generate top words from these topics. In this process LDA is
+    used in combination with algorithms which generate document-term-
+    matrixes, like :meth:`count` or :meth:`tfidf`
+
+    LDA can directly handle sparse input, so when calling truncatedSVD on a
```
jbesomi, line 989: truncatedSVD??

texthero/representation.py
```diff
@@ -947,22 +893,730 @@ def meanshift(
...
+    DocumentTermDF, the advantage of sparseness is kept.
```
jbesomi, line 990: We haven't mentioned what's a DocumentTermDF; i.e the bottomline is that we need to make clear what's the _kind_ of input and output

texthero/representation.py
```diff
@@ -947,22 +893,730 @@ def meanshift(
...
+
+    Parameters
+    ----------
+    s : pd.Series (VectorSeries) or MultiIndex Sparse DataFrame (DocumentTermDF)
```
jbesomi, line 994: Same comments as #156

texthero/representation.py
```diff
@@ -947,22 +893,730 @@ def meanshift(
...
+
+    n_components : int, default is 2.
```
jbesomi, line 996: default is 2. vs (default: 10) [below]; we need to stick to a single approach; TODO we should also add this under CONTRIBUTING.md

texthero/representation.py
```diff
@@ -947,22 +893,730 @@ def meanshift(
...
+        Number of components to keep (dimensionality of output vectors).
+        When using truncatedSVD for Topic Modelling, this needs to be
+        the number of topics.
+
+    max_iter : int, optional (default: 10)
+        The maximum number of iterations.
```
jbesomi, line 1002: Of what? Explain further

texthero/representation.py
```diff
@@ -947,22 +893,730 @@ def meanshift(
...
+    cell.
+
+    Examples
+    --------
+    >>> import texthero as hero
+    >>> import pandas as pd
+    >>> s = pd.Series(["Football, Sports, Soccer", "Music, Violin, Orchestra",
+    ...                "Football, Music"])                
+    >>> s = s.pipe(hero.clean).pipe(hero.tokenize).pipe(hero.term_frequency)
+    >>> hero.lda(s, random_state=42) # doctest: +SKIP
```
jbesomi, line 1020: We should not skip the doctest

texthero/representation.py
```diff
@@ -947,22 +893,730 @@ def meanshift(
...
+    max_iter : int, optional (default: 10)
+        The maximum number of iterations.
+
+    random_state : int, default=None
+        Determines the random number generator. Pass an int for reproducible
+        results across multiple function calls.
+
+    Returns
+    -------
+    Pandas Series with the vector calculated by LDA for the document in every
```
jbesomi, line 1010: Is not more appropriate to return a DataFrame? + The Pandas Series is of which type?

jbesomi:
For now, reviewed only `lda`, see comments below

texthero/representation.py
henrifroese: will change

texthero/representation.py
henrifroese: will change

texthero/representation.py
henrifroese: oops, will change

texthero/representation.py
henrifroese: Yes, we'll need to go through all PRs that depend upon #156 once more. But we'll only do this after it is merged so we don't have to change everything several times :pretzel:

texthero/representation.py
henrifroese: Yes, see above; we'll go through this after 156 is merged.

texthero/representation.py
henrifroese: Yes that's right. Will change all defaults to same value and add this to CONTRIBUTING later.

texthero/representation.py
henrifroese: will do

texthero/representation.py
henrifroese: We need to skip this because of floating-point-errors /-differences on Mac/Windows as travis will fail (it'll only work locally). But we're also testing LDA of course in test_representation, so I think it's okay

texthero/representation.py
henrifroese: Changed the docstring. I don't think it's more appropriate to return a DataFrame as we said we'll only do this for the sparse stuff and things that are not meant for integration into a dataframe (e.g. the topic-term-matrix from `topic_matrices`). For pca/tsne/... we want the VectorSeries output as it is not directly interpretable (so we don't profit from the column names in a DataFrame) and can be put into a DataFrame.

texthero/representation.py
henrifroese: See below; changed currently to `Union[VectorSeries, pd.DataFrame]`

henrifroese:
Thanks for the review! As I commented above, we'll have to go through this again anyway once #156 is merged :pray: .

texthero/representation.py
jbesomi: Have you tried solving the floating error points problem using [doctest ellipsis](https://www.kite.com/python/examples/2699/doctest-ignore-some-output-in-a-docstring-test-using-the-ellipsis-option)?

jbesomi:
#156 has been merged; can you please go through it again?

mk2510:
we also have updated this branch, so it is now sourced from the master 🥳

kepler:
This would be a very useful feature. Any pending blockers or any expected date for merging and releasing?

jbesomi:
Hey @kepler
Yes, the plan is to merge this PR soon. But first, the idea is to release a new version with the HeroSeries (to introduce and explain the concept). After that, we will be able to merge this one.

The remaining step for the HeroSeries is:
1) make sure each function makes correctly use of the HeroSeries and test
2) finalize the documentation for the HeroSeries (#117, #118, #135)

bcornet1:
Hello, do you have any news on this topic or when it will be released? Thanks :)

havardl:
Hi, is there any news on when this PR will be implemented?

Issue #155, opened 2020-08-20
Missing `fit` feature


jbesomi:
In any ML task, the assumption is that the test data are not available during training and just available in the prediction phase.

Assume someone wants to categorize reviews using `tfidf` + Naive Bayes. The required step would be the following ones:

1. Split train and test
1. Fit `tfidf` on the train part and generate (the transform part in scikit-learn) the `tfidf` values on the train part
1. Train the model
1. Generate the tf-idf values on the test part, this time using the already fitted model

The problem is that with the current implementation we don't have any state (and that brings also many advantages such as simplicity). The `tfidf` functions do not return any already fitted model, rather the already transformed values.

We need to take a clear position wrt to this point. Having the exact same approach as scikit-learn would not probably make sense, still, we need to consider this fact. Opinions?

jbesomi:
In sklearn and similar toolkit we generally `model.fit_transform` on train data and only `model.transform` on text data. In Texthero we cannot do that as we don't have any model object (but that's fine somehow, we just need to have a clear position on that)

mk2510:
so far I didn't start thinking about the missing fit feature, as fit is in general called on models, which can be fitted to your dataset (at least from my experience so far). 
In those cases, where we work with models - which can be fitted, like in 'pca' or 'kmeans' - I think, we can mention it on the getting started page, that we call everytime fit_transform and don't provide the option to store the fitted pca model. But the way the API is designed, the user probably wouldn't assume it anyways, I guess 😬 🙈 :octocat:

jbesomi:
Agree. Yes, we can mention that in the (future) FAQ page.

harshraj-wadhwani:
@mk2510 Any plans to add `fit` method ? 
Or is there any way by which I can fit on train data, save to pickle file, and transform on test data ?

PR #167, opened 2020-08-26
Implement train_test_split

mk2510:
Implemented the method *train_test_split* to prepare a DataFrame for machine learning algorithms, by dividing it into 2-3 smaller DataFrames, which represent the train, test (and optionally validation) sets and additional test. The idea is that users are working with a DataFrame that has columns `text, pca, ..., X, Y` where columns `X, Y` are columns with input and labels/output (e.g. X might be a bert embedding of the text, Y might be class targets and users want to train an ML model to predict the labels). Most users currently use `sklearn.train_test_split` which we also internally use in the new function, but with this function we
- provide a unified interface for users (don't have to import sklearn for their ML preprocessing pipeline)
-  extend the sklearn function to split into train-test-validation (sklearn can only split into train-test)

## Overview
```python
def train_test_split(df,train=0.0,test=0.0,val=0.0,class_balance=None,shuffle=True,random_state=None) -> 
Union[
    Tuple[pd.DataFrame, pd.DataFrame], Tuple[pd.DataFrame, pd.DataFrame, pd.DataFrame]
]:
```
Here the method takes:
-  a DataFrame that we want to split up into training, testing, and optionally validation.
- train=0.0,test=0.0,val=0.0, where, when all are set to 0.0, then the proportion will be 70% training set and 30% test set. If just validation is set, then we will return an error, as no training and test sizes are defined. If ints instead of floats are given, it will interpret this as the absolute value (e.g. train=200 -> train DF will have 200 documents)
- class_balance: if the data is already labeled and we want to keep the label proportions in the subsets (same as sklearn `stratify`), we pass a Series with the labels here (see example below)
- shuffle, which shuffles the dataset
- random_state, to create deterministic results

The method will return 2 or 3 DataFrames in a Tuple, with at the first 🥇 position the train_set_DataFrame, at the second 🥈 place the test_set_DataFrame and if validation was set to a different value than 0.0 it will return on the third 🥉 place the validation_set_DataFrame.

## Example

```python

>>> import texthero as hero
>>> import pandas as pd
>>> df = pd.read_csv(
...    "https://raw.githubusercontent.com/jbesomi/texthero/master/dataset/bbcsport.csv"
... ) # doctest: +SKIP
>>> train_df, test_df = hero.train_test_split(
...                  df, class_balance = df["topic"], random_state = 42, train=0.8
... ) # doctest: +SKIP
>>> len(train_df) # doctest: +SKIP
589
>>> len(test_df) # doctest: +SKIP
148
>>> # to check wether we have a equal class distribution (in this case a class is a topic)
>>> # we will look at the percentage of each topic in the DataFrame
>>> train_df["topic"].value_counts() / train_df["topic"].value_counts().sum() # doctest: +SKIP
football     0.359932
rugby        0.198642
cricket      0.168081
athletics    0.137521
tennis       0.135823
Name: topic, dtype: float64
>>> # and now the same for the test dataset
>>> test_df["topic"].value_counts() / test_df["topic"].value_counts().sum() # doctest: +SKIP
football     0.358108
rugby        0.202703
cricket      0.168919
athletics    0.135135
tennis       0.135135
Name: topic, dtype: float64
```

### DIFF
--- a/tests/test_preprocessing.py
+++ b/tests/test_preprocessing.py
@@ -381,3 +381,78 @@ class TestPreprocessing(PandasTestCase):
         s_true = pd.Series("Hi  , we will remove you")
 
         self.assertEqual(preprocessing.remove_hashtags(s), s_true)
+
+    """
+    Test train_test_split
+    """
+
+    def test_train_test_split(self):
+        df = pd.DataFrame(["Here", "hello", "there", "Test"])
+        train_set, test_set = preprocessing.train_test_split(
+            df, random_state=42, test=0.5
+        )
+        df_train_true = pd.DataFrame(["Here", "there"], index=[0, 2])
+        df_test_true = pd.DataFrame(["hello", "Test"], index=[1, 3])
+        pd.testing.assert_frame_equal(train_set, df_train_true)
+        pd.testing.assert_frame_equal(test_set, df_test_true)
+
+    def test_train_test_split_wrong_param(self):
+        df = pd.DataFrame(["Here", "hello", "there", "Test"])
+        self.assertRaises(
+            ValueError, preprocessing.train_test_split, df=df, random_state=42, val=0.4
+        )
+
+    def test_train_test_split_wrong_param_2(self):
+        df = pd.DataFrame(["Here", "hello", "there", "Test"])
+        self.assertRaises(
+            ValueError,
+            preprocessing.train_test_split,
+            df=df,
+            random_state=42,
+            train=5,
+            test=3,
+        )
+
+    def test_train_test_val_split(self):
+        df = pd.DataFrame(["Here", "hello", "there", "Test"])
+        train_set, test_set, val_set = preprocessing.train_test_split(
+            df, random_state=42, test=0.5, val=0.25
+        )
+        df_train_true = pd.DataFrame(["there"], index=[2])
+        df_test_true = pd.DataFrame(["Test", "Here"], index=[3, 0])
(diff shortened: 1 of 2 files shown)


jbesomi:
As Texthero cannot really offer ML capabilities (see #155 for instance), I don't really see where users can make use of this function as they would need anyway to import and use scikit-learn ...

Also, split into train-test-dev is quite straightforward:

```
X_train, y_train, X_test, y_test = train_test_split(df[['columns', ...,]], df['target], ...)
X_val, y_val, X_test, y_test = train_test_split(X_test, y_test, ...)
```

Not sure users would want to read the docstring of a new function from a library not designed for these kinds of ML operations ...

Sorry for the negative review! :( 👍 

Would be happy to see your motivations though!

mk2510:
> Would be happy to see your motivations though!

we saw three main advantages in this implementation
1. the users don't have to import sklearn, when they do machine learning. They can preprocess the text with Texthero, split it with Texthero, and then dump it into TensorFlow.  This way the user saves to import one additional library
2. it makes the code of the user shorter as it saves one line
3. we have the option to keep the class balance even for the case with validation, so the user doesn't have to create his own solution, so he doesn't have to google it.

But no worries if it's not gonna be merged, as we do this project for fun 🪂 🥇 ⛽

jbesomi:
Closing this as this is out-of-scope, at least for now. Thank you for the contribution nonetheless! 👌

PR #165, opened 2020-08-25
Fix Term_Frequency

mk2510:
The `term_frequency` function is currently implemented incorrectly. Example:

```python
>>> s = pd.Series(["Text Text of doc one", "Text of of doc two", "Aha hi bnd one"]).pipe(hero.tokenize)
>>> s
0    [Text, Text, of, doc, one]
1      [Text, of, of, doc, two]
2           [Aha, hi, bnd, one]
dtype: object
```

Old implementation of term frequency returns:
```python
  term_frequency                                                                      
             Aha      Text       bnd       doc        hi        of       one       two
0       0.000000  0.142857  0.000000  0.071429  0.000000  0.071429  0.071429  0.000000
1       0.000000  0.071429  0.000000  0.071429  0.000000  0.142857  0.000000  0.071429
2       0.071429  0.000000  0.071429  0.000000  0.071429  0.000000  0.071429  0.000000
```

New version returns:

```python
  term_frequency                                      
            Aha   Text  bnd  doc    hi   of   one  two
0           0.00  0.4  0.00  0.2  0.00  0.2  0.20  0.0
1           0.00  0.2  0.00  0.2  0.00  0.4  0.00  0.2
2           0.25  0.0  0.25  0.0  0.25  0.0  0.25  0.0
```

We can see that the old version simply did `count` for each document and then divided each count value by _the total number of terms across all documents_ (the `.sum()` in the old code). That does not make sense for `term_frequency`. The new version will also do count, but then divide the count values of each document by the number of terms _in that document_ to produce the correct term frequencies. Of course, that's the same as L1-Normalizing every row, so that's how we implemented it 🚥 .

**Note**: only so many lines changed as this builds upon the DocumentTermDF (see #156)

PR #169, opened 2020-08-26
Implement filter_extremes

henrifroese:
We add a new function `hero.filter_extremes(s: TokenSeries, max_words=None, min_df=1, max_df=1.0)` to remove words from all documents that are above or below a document frequency threshold; additionally only keep max_words many words. Naming from [gensim's similar function here](https://radimrehurek.com/gensim/corpora/dictionary.html#gensim.corpora.dictionary.Dictionary.filter_extremes).

## Excerpt from docstring to explain functionality:

Decrease the size of your documents by
filtering out words by their frequency.

It is often useful to reduce the size of your dataset
by dropping words in order to
reduce noise and improve performance.
This function removes all words/tokens from
all documents where the
document frequency (=number of documents a term appears in) is
-  below min_df
- above max_df.

When min_df or max_df is an integer, then document frequency
is the absolute number of documents that a term
appears in. When it's a float, it is the
proportion of documents a term appears in.

Additionally, only max_words many words are kept.

**Parameters**

max_words : int, default to None
The maximum number of words/tokens that
are kept, according to term frequency descending.
If None, will consider all features.

min_df : int or float, default to 1
Remove words that have a document frequency
lower than min_df. If float, it represents a
proportion of documents, integer absolute counts.

max_df : int or float, default to 1
Remove words that have a document frequency
higher than max_df. If float, it represents a
proportion of documents, integer absolute counts.


henrifroese:
Note: Black (our formatter) just rolled out V20.8b1 3 days ago. This creates errors with our ./tests.sh in preprocessing because of whitespace. Will investigate this further but atm we set the black version in `.travis.yml` and `setup.cfg` to the last working version (19.10b1).

jbesomi:
Thanks. Will review once the previous PRs are merged

jbesomi:
Waiting for #162 to be merged + will need to conflicts change (and will simplify the code).

mk2510:
this branch depends on branch PR #165 So will be set ready to review when this is merged

PR #174, opened 2020-09-04
Fix type checks to not fail if first value in series is NaN/missing.

henrifroese:
Currently, when using our `InputSeries` decorator, it always throws an error if the first Series value is NaN. This of course makes no sense for users.

Example:
```python
>>> import pandas as pd
>>> import texthero as hero
>>> s = pd.Series([pd.NA, "test"])
>>> hero.clean(s)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/henri/Prog/SummerOfCode/texthero/texthero/texthero/_types.py", line 233, in wrapper
    allowed_hero_series_type.check_series(s)
  File "/home/henri/Prog/SummerOfCode/texthero/texthero/texthero/_types.py", line 124, in check_series
    raise TypeError(error_string)
TypeError: The input Series should consist only of strings in every cell. See help(hero.HeroSeries) for more information.
```

We fix this by finding the first non-NaN value now and checking this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ jobs:
       env: PATH=/c/Python38:/c/Python38/Scripts:$PATH
 install: 
     - pip3 install --upgrade pip  # all three OSes agree about 'pip3'
-    - pip3 install black
+    - pip3 install black==19.10b0
     - pip3 install ".[dev]" .
 # 'python' points to Python 2.7 on macOS but points to Python 3.8 on Linux and Windows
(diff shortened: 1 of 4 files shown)


henrifroese:
Accidentally pushed the nlp tutorial on here, so had to force-push one commit back. It's now back at the correct commit (0cca7d9) :upside_down_face:

jbesomi:
Good catch, thanks! 🎉 

When reviewing, I noticed a couple of other issues (see #180 ) that preferably need to be dealt with separately (one PR for each issue). Would you mind looking into that, please?

PR #175, opened 2020-09-05
Visualisation getting started

mk2510:
created a short tutorial, which will introduce the user to the two visualization functions top_words 🔝  and word cloud. ☁️  I left out the scatterplot function on purpose, as I couldn't make that one fit to the story as well, without introducing a bunch of representation functions. Therefore this functionality will be introduced in the representation tutorial. Content-wise it also makes more sense to explain the user, what a scatterplot is, after introducing PCA, so the user knows, what he is plotting. The topic modeling tutorials will follow in a separate tutorial after the PRs for topic modeling are approved. 🥇


mk2510:
we went through this PR again and this is ready to review / to be merged from our side 🍶 🧄

Issue #171, opened 2020-08-29
Doctests fail with new Black version.


henrifroese:
As we could see in #169 and #170 , with the [new release of black](https://github.com/psf/black/releases) three days ago, many of our doctests fail. We opened an issue [here at Black's Github](https://github.com/psf/black/issues/1654). Currently, when we want to push something new to Texthero, Travis CI will fail. Current hotfix is to use the last version before the release by setting in `.travis.yml`: `- pip3 install black==19.10b0` and in `setup.cfg`: `black==19.10b0`.

jbesomi:
@henrifroese thanks! Can I close this issue?

henrifroese:
yes should be good

PR #168, opened 2020-08-26
Implement hero.describe(s), Closes #166

henrifroese:
Straightforward implementation of description in #166 . 

Example:

```python
>>> import texthero as hero
>>> import pandas as pd
>>> df = pd.read_csv("https://raw.githubusercontent.com/jbesomi/texthero/master/dataset/bbcsport.csv")
>>> hero.describe(df["text"], df["topic"])
                                                                                              Value
number of documents                                                                             737
number of unique documents                                                                      727
number of missing documents                                                                       0
most common words                                          [the, to, a, in, and, of, for, ", I, is]
most common words excluding stopwords             [said, first, england, game, one, year, two, w...
average document length                                                                     387.803
length of shortest document                                                                     119
length of longest document                                                                     1855
standard deviation of document lengths                                                      210.728
25th percentile document lengths                                                                241
50th percentile document lengths                                                                340
75th percentile document lengths                                                                494
label distribution                     football                                            0.359566
                                       rugby                                               0.199457
                                       cricket                                              0.16825
                                       athletics                                           0.137042
                                       tennis                                              0.135685

```

Screenshot of pretty-printed output from Google Colab:

Screenshot from 2020-08-26 15-41-24

PR #178, opened 2020-09-06
Visualize describe with plots

mk2510:
This new function in the visualization module allows users to easily visualize the describe function. We belive, that this will be often use, to display the key characteristics of the DataFrame. 🥉 

This function simply calls the describe functions and then with plotly those plots will be generated and displayed in a new browser window. As it is a lot of code to display them, we belive, that it will be quite an enrichment to the library 🏑 

Note: travis/setup changed because of #171 . This is branched straight from PR #168.

[This](https://colab.research.google.com/drive/1VERIt61d8u6VtO9E85tSbvwArVcgG1Tp?usp=sharing) is a short colab demo of the new function 🤖


texthero/visualization.py
```diff
@@ -306,3 +308,123 @@ def top_words(s: TextSeries, normalize=False) -> pd.Series:
         .explode()  # one word for each line
         .value_counts(normalize=normalize)
     )
+
+
+def visualize_describe(df, text_col_name="text", labels_col_name="topic"):
```
jbesomi, line 313: `labels_col_name="topic"` not sure it make sense to have as default value "topic". Moreover, there might be multiple non-text columns.

Going rither, as we are interested in giving users information regarding text-columns, I would limit this function to handle text-columns. I.e replace `text_col_name ` with `text_cols_names` (or similar) and accept both a list of string or just a string (or only a list of string, as you wish.). I suggest to drop `labels_col_name` but I'm open to discussion and to see your point of view

jbesomi:
That's beautiful! 😍 

- What does it happen if you run `visualize_describe` on an empty DF? (we should test that behavior for all heroes functions ...)
- (opinionable comment) What if we have all graphs in a single column? That's might be better for visualization on narrow displays

texthero/visualization.py
henrifroese: see comment below

henrifroese:
> What does it happen if you run visualize_describe on an empty DF? (we should test that behavior for all heroes functions ...)

Currently we get an error but that's just because of the default arguments (see below). Agree we should test that behavior for all hero functions, probably in a separate PR in the future.

> (opinionable comment) What if we have all graphs in a single column? That's might be better for visualization on narrow displays

We tried it out and think it looks good as-is on 13 inch screens (which is the standard "small" laptop size).

**From your review**

> labels_col_name="topic" not sure it make sense to have as default value "topic". Moreover, there might be multiple non-text columns. ...

Yes, the default arguments really make no sense, we just used them for testing purposes and forgot to remove them :facepalm: :man_facepalming: :woman_facepalming: 

Additionally, not sure what you mean / how we would integrate several text columns? We thought that there are two main scenarios:

1. User has text in a column -> he uses our function without the labels stuff, everything's great

2. User has text in a column and labels in a column (e.g. to train a classifier) -> he uses our function with the labels stuff, everything's great

Additionally just noticed that there isn't a docstring (totally forgot this). Will convert this to draft and address all concerns + add docstring. Will update here and post screenshots when it's done.

henrifroese:
Now changed the function signature to be in line with `hero.describe` (which really makes sense as this is a visualization of hero.describe). Additionally, default arguments are changed.

We noticed that this function _cannot_ handle empty input because of `hero.tokenize` cannot handle empty input. This is probably a low-priority bug as of course empty input does not make any sense whatsoever.

mk2510:
we just went through this PR again and it is ready to be reviewed / to be merged 🔢 🧶

PR #181, opened 2020-09-08
Add infer test cases for test indexes

AlfredWGA:
This PR add codes to generate test cases for each function according their input HeroSeries. Only functions with one HeroSeries input will be generated with test case. Exceptions can be manually added, for example,

```
test_case_exceptions = {}
for case in (
    test_cases_nlp
    + test_cases_preprocessing
    + test_cases_representation
    + test_cases_visualization
):
    test_case_exceptions[case[0]] = case
```
Then functions within `test_case_exceptions` will not be given a generated test case.
To omit some functions for testing, put them under `func_white_list`,
```
func_white_list = set(
    [s for s in inspect.getmembers(visualization, inspect.isfunction)]
)
```

### DIFF
--- a/tests/test_indexes.py
+++ b/tests/test_indexes.py
@@ -1,10 +1,14 @@
 import pandas as pd
 from texthero import nlp, visualization, preprocessing, representation
+import texthero as hero
 
 from . import PandasTestCase
 import unittest
 import string
 from parameterized import parameterized
+import importlib
+import inspect
+import numpy as np
 
 
 # Define valid inputs for different functions.
@@ -12,6 +16,13 @@ s_text = pd.Series(["Test"], index=[5])
 s_tokenized_lists = pd.Series([["Test", "Test2"], ["Test3"]], index=[5, 6])
 s_numeric = pd.Series([5.0], index=[5])
 s_numeric_lists = pd.Series([[5.0, 5.0], [6.0, 6.0]], index=[5, 6])
+
+# Define the valid input for each HeroSeries type
+valid_inputs = {
+    "TokenSeries": s_tokenized_lists,
+    "TextSeries": s_text,
+    "VectorSeries": s_numeric_lists,
+}
 
 # Define all test cases. Every test case is a list
 # of [name of test case, function to test, tuple of valid input for the function].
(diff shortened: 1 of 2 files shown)


jbesomi:
Thanks, great! 🎉 🎉 

The file still contains a lot of commented lines that appear to be unnecessary. Can you please clean it a bit? Also, we might need to update the comments by explaining the new changes as well as removing the unnecessary ones, can you please have a detailed look at that too?

Regards,

henrifroese:
Looks great! I think that if some other developer now adds a function, he might get a weird error if this file automatically generates a test case and it maybe fails (e.g. because it needs additional arguments and it's not 100% obvious to everyone where to implement that). That's why I think it would be great if you could add some explanation for other developers in the file. E.g.

```python
"""
The tests below Blabla...
"""
```

That's also what we did in `_types`, hoping it will help others fully understand the code.

AlfredWGA:
@jbesomi @henrifroese Just add some comments for explanation and delete the old ones. 👌

tests/test_indexes.py
```diff
@@ -81,13 +86,56 @@
...
     test_cases_nlp
     + test_cases_preprocessing
     + test_cases_representation
     + test_cases_visualization
+):
+    test_case_custom[case[0]] = case
+
+
+# Put functions' name into white list if you want to omit them
+# func_white_list = {
```
henrifroese, line 101: Remove this and the next few lines? (The comment)

henrifroese (commented):
Looks good to me overall 👍

tests/test_indexes.py
AlfredWGA: Done!

tests/test_indexes.py
```diff
@@ -1,58 +1,63 @@
...
+}
+
+# Specify your custom test cases here (functions that
+# has multiple arguments, doesn't accpet HeroSeries, etc.)
+test_cases_nlp = []
 
 test_cases_preprocessing = [
-    ["fillna", preprocessing.fillna, (s_text,)],
-    ["lowercase", preprocessing.lowercase, (s_text,)],
     ["replace_digits", preprocessing.replace_digits, (s_text, "")],
```
jbesomi, line 55: Can we generalize over these functions as well and avoid to define the custom test manually? Rewriting the file as you are proposing add more complexity for the developer, I can see it as worth it in case the custom functions that need to be manually defined are kept to the very minimum

AlfredWGA: Can I assume that any functions with default params can be auto-tested?

jbesomi: Seems reasonable, opinions?

AlfredWGA: Then I think the simplest way might be regulating the function definition and add default params whenever possible, so the default params can be taken as a default test case.

jbesomi:
Hey @AlfredWGA, what's the current status for this PR?

AlfredWGA:
> Hey @AlfredWGA, what's the current status for this PR?

I haven't come up with any idea on how to further avoid defining test case manually. Maybe leave it to another PR in the future?

PR #185, opened 2020-09-13
Add and enforce best practices

henrifroese:
This PR adds new best practices to CONTRIBUTING.md and enforces them in the code:

1. Docstring for default values (now e.g. `x : int, optional, default=2`)
1. Selection of default argument values (try to use defaults of the underlying library called, else look at other similar functions; finally explain if something else is used)
1. American English (instead of British English)


jbesomi:
Amazing, thanks! 🎉 🎉

PR #176, opened 2020-09-06
Add NLP Tutorial

henrifroese:
This builds upon

- #174 (the tutorial does not work otherwise - really important)
- #171 

TODO

- add references at the end to the other tutorials

Issue #173, opened 2020-09-03
Add Lemmatization


henrifroese:
Lemmatization can be thought of as a more advanced `stemming` that we already have in the preprocessing module. You can read about it e.g. [here](https://en.m.wikipedia.org/wiki/Lemmatisation). Implementation should be done with spaCy.

# ToDo

Implement a function `hero.lemmatize(s: TokenSeries)` (or mayber rather TextSeries?).  Using spaCy this should be fairly straightforward. It should go into the NLP module and probably look very similar to the other spacy-based functions there.

Just comment below if you want to work on this and/or have any questions. I think this is a good first issue for new contributors.

adrianhosak:
Hi, I would like to work on this issue.

jbesomi:
That's great! 

Regarding the function signature, not sure which one is better (maybe even accept both Series type might make sense?). But, it will depend on the implementation requirements. Probably with spaCy we will have to first tokenize the text and then take the lemma with `token.lemma_` or something similar ... so maybe we need to pass a `TextSeries`? (as we will need to pass the sentence into `nlp(sentence)`)

Also, independently of the choice, both `stem` (should we rename it to `stemming`) and `lemming` should probably have the same signature.

Regards,

adrianhosak:
Using the `token.lemma_` attribute, a straightforward implementation would be something like this:
   ```
    lemmatized_docs = []

    nlp = spacy.load("en_core_web_sm", disable=["ner"])

    for doc in nlp.pipe(s.astype("unicode").values, batch_size=32):
        lemmatized_docs.append(" ".join([word.lemma_ for word in doc]))

    return pd.Series(lemmatized_docs, index=s.index)
```
which is very similar to other functions in the nlp.py folder. And the nlp does require a string as input and not a token so I think it makes sense to make it a TextSeries signature. I set batch_size to 32 here because thats how it is in all the other functions, but I don't know if and why that's a good choice.

What I found to be a problem is that `token.lemma_` seems to replace any pronouns with `-PRON-`. For example:
```
>>>  s = pd.Series(["My name is Adrian"])
>>>  hero.lemmatize(s)
0    -PRON- name be adrian
dtype: object
```
We don't really want that, right? Maybe we can use `hero.pos_tag` first to check wether a token is a pronoun. What do you guys think?

henrifroese:
I agree we probably do not want that. Have you found an alternative solution? :3rd_place_medal:

AlfredWGA:
[...]

PR #186, opened 2020-09-13
Move stem from preprocessing to nlp

henrifroese:
We believe the stem function belongs to the NLP module and not the preprocessing module (together with lemmatization that will be implemented in #173 ). Of course it _can_ make sense to use it in preprocessing (and we should explain that in the NLP tutorial in #176 , but in our opinion it's more "Natural Language Processing" than "Preprocessing" so it should be moved. 

Of course, no code is changed in this PR, only moved.


jbesomi:
Very much agree!

Thanks. 🎉

Issue #182, opened 2020-09-10
fillna should support customized fill content


monkdou0:
texthero.preprocessing.fillna()  only replace the nan with writespace.
It should support customized fill content

henrifroese:
Makes sense. This looks like a good first issue! Do you want to work on this? @monkdou0

monkdou0:
Of course.
I will try and then give the PR

k0pernicus:
I think this issue has been solved in #190, no?

jbesomi:
Right, thank you @k0pernicus!

PR #190, opened 2020-09-25
Optional string in fillna

Andre-Sacilotti:
Hey, i added an optional argument in fillna, for those, like me, who need to fill missing values with an custom string.

Example:

```python
>>> hero.fillna(s, "Custom String!!")
0                I'm
1    Custom String!!
2    Custom String!!
3             You're
dtype: object
```

### DIFF
--- a/texthero/preprocessing.py
+++ b/texthero/preprocessing.py
@@ -24,7 +24,7 @@ warnings.filterwarnings(action="ignore", category=UserWarning, module="gensim")
 
 
 @InputSeries(TextSeries)
-def fillna(s: TextSeries, replace_string = "") -> TextSeries:
+def fillna(s: TextSeries, replace_string="") -> TextSeries:
     """
     Replaces not assigned values with empty or given string.
 
--- a/texthero/preprocessing.py
+++ b/texthero/preprocessing.py
@@ -24,24 +24,31 @@ warnings.filterwarnings(action="ignore", category=UserWarning, module="gensim")
 
 
 @InputSeries(TextSeries)
-def fillna(s: TextSeries) -> TextSeries:
+def fillna(s: TextSeries, replace_string = "") -> TextSeries:
     """
-    Replaces not assigned values with empty string.
-
+    Replaces not assigned values with empty or given string.
 
     Examples
     --------
     >>> import texthero as hero
     >>> import pandas as pd
+    >>> import numpy as np
     >>> s = pd.Series(["I'm", np.NaN, pd.NA, "You're"])
     >>> hero.fillna(s)
     0       I'm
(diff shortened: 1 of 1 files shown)


Andre-Sacilotti:
Closes #182

texthero/preprocessing.py
```diff
@@ -24,32 +24,39 @@
 
 
 @InputSeries(TextSeries)
-def fillna(s: TextSeries) -> TextSeries:
+def fillna(s: TextSeries, replace_string: Optional[str] = "") -> TextSeries:
```
mk2510, line 27: I wouldn't do the optional type here, either the "" or the typing with Optional[str] but not both, as we don't gain any informations from that.

texthero/preprocessing.py
```diff
@@ -24,32 +24,39 @@
...
     """
-    Replaces not assigned values with empty string.
+    Replaces not assigned values with empty (Default) or given string.
```
mk2510, line 29: the default here as well is too much in my opinion. The user will probably recognise, that "" is an empty string

mk2510 (commented):
Looks great overall 🥳 
I just had two small comments for the signature of the function and the docstring. But I like this addition!

texthero/preprocessing.py
Andre-Sacilotti: Hey, thanks for the comments, this help us learn more. I will commit the changes

henrifroese:
[Here](https://github.com/jbesomi/texthero/pull/190/files) we can see that your changes change a lot of "invisible" stuff (whitespace) in the whole preprocessing file, which of course we cannot accept as this leads to doctests failing. Not sure which Black version you are using? You can see [here](https://github.com/jbesomi/texthero/issues/171) that we sadly have problems with the newest version; that might be the issue 😿

Andre-Sacilotti:
> [Here](https://github.com/jbesomi/texthero/pull/190/files) we can see that your changes change a lot of "invisible" stuff (whitespace) in the whole preprocessing file, which of course we cannot accept as this leads to doctests failing. Not sure which Black version you are using? You can see [here](https://github.com/jbesomi/texthero/issues/171) that we sadly have problems with the newest version; that might be the issue crying_cat_face

oh, i just saw the travis ci build failed, i'm using black-19.10b0, i'm trying to resolve this

Andre-Sacilotti:
Okay, i think its working right now. Any tips or comments will be very helpful.

jbesomi:
Hey @Andre-Sacilotti, looks great! Thank you!

And welcome to Texthero as a contributor, you can open a new PR and add your name in the README.md file! 👍

PR #194, opened 2020-10-16
Lazy load stopwords module to prevent downloading when it's not needed

### LINKED ISSUE #193: prevent texthero from downloading stopwords
When importing texthero, by default the stopwords get downloaded. 
you might not always use stopwords in your job and this can get you into trouble. What about downloading the stopwords only when we need them? 

For the rest great work guys :)

hugoabonizio:
Import `stopwords` module inside `replace_stopwords` function, so the data is downloaded only when it's necessary.

Fixes #193

### DIFF
--- a/texthero/nlp.py
+++ b/texthero/nlp.py
@@ -4,9 +4,17 @@ The texthero.nlp module supports common NLP tasks such as named_entities, noun_c
 
 import spacy
 import pandas as pd
-import en_core_web_sm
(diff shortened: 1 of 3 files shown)


hugoabonizio:
I also moved the spaCy model downloading to `nlp.py` file.

jbesomi:
Thank you, Hugo! And welcome to Texthero's contributors family. You can open a new PR and add your name under the contributors in the homepage README.md :)

PR #206, opened 2021-04-06
Investigate build and cicd issues

henrifroese:
There are problems with dependencies (spacy and gensim especially) leading to failing builds.


jbesomi:
I see, we should probably define the variable name for each library to avoid these kinds of issues also in the future.

henrifroese:
@jbesomi build issues are now fixed with this pull request, can be merged.

- set spaCy to <3.0.0
- set gensim>=3.6.0,<4.0
- change `class HeroTypes(pd.Series, pd.DataFrame)` to `class HeroTypes` without base classes. The base classes are not needed and give a method resolution order error for newer python version
- test for both possibilities of equality for categorical results: e.g. when we test kmeans and expect result `[1,0,1,0]` and kmeans gives us the result `[0,1,0,1]` (which as 1 and 0 are just categories is functionally the same), we also accept that result

jbesomi:
Very good job! 🎉  These parts were tricky yet very important! Thanks a lot, Henri 👍 :)

PR #235, opened 2023-08-29
Fix failing CI/tests Workflows

jbesomi:


jbesomi:
@RobertRosca FYI: this PR removes test for python 3.12-dev

